PR #3, opened 2016-06-21
fix cookies for local environment

danielepolencic:
At the moment no session cookie is dropped when the app is started locally (`http://localhost:8080`).

### DIFF
--- a/lib/defaults.js
+++ b/lib/defaults.js
@@ -29,6 +29,7 @@ const defaults = {
     host: process.env.REDIS_HOST || '127.0.0.1'
   },
   session: {
+    name: 'hof.sid',
     ttl: process.env.SESSION_TTL || 1800,
     secret: process.env.SESSION_SECRET || 'changethis'
   },
--- a/lib/sessions.js
+++ b/lib/sessions.js
@@ -44,18 +44,17 @@ module.exports = (app, config) => {
   app.use(cookieParser(config.session.secret, {
     path: '/',
     httpOnly: true,
-    secure: true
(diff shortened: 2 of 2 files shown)


lib/sessions.js
```diff
@@ -44,18 +44,16 @@ module.exports = (app, config) => {
...
     httpOnly: true,
-    secure: true
+    secure: config.protocol === 'https'
   }));
 
-  const sessionOpts = Object.assign({}, {
+  const sessionOpts = Object.assign({
     store,
-    cookie: {secure: true},
-    key: '',
```
JoeChapman: I rather like the idea of having a custom cookie name. Should we default it something other than '', but make it configurable?

danielepolencic: I couldn't find documentation for the `key` property?

JoeChapman: Looks as if it might now be called `name` https://github.com/expressjs/session/blob/master/index.js#L89

danielepolencic: `name: 'hof.sid'` ?

danielepolencic:
the cookie name was updated.

ping @JoeChapman

lib/defaults.js
```diff
@@ -29,6 +29,7 @@ const defaults = {
     host: process.env.REDIS_HOST || '127.0.0.1'
   },
   session: {
+    name: 'hof.sid',
```
JoeChapman: Is this something that should be optionally overridden by the user so that they can name the cookie something relevant to their service?

Issue #4, opened 2016-06-21
config is lost when start up is delayed


danielepolencic:
When `startOnInitialise` is `false` when the interface is `start`ed, the same config has to be passed to the method otherwise it will fail:

```javascript
const path = require('path');
const bootstrap = require('hof-bootstrap');

bootstrap({
  startOnInitialise: false,
  routes: [{
    fields: path.resolve(__dirname, 'fields'),
    views: path.resolve(__dirname, 'views'),
    steps: {
      '/one': {
        next: '/two',
        fields: ['fname', 'lname']
      },
      '/two': {
        template: 'two'
      }
    }
  }]
}).then(bootstrap => {
  bootstrap.start(); // silently fails
  bootstrap.start({startOnInitialise: true}); // silently fails
  bootstrap.start({startOnInitialise: true, protocol: 'http', port: '8080', host: '0.0.0.0'}); // finally works
});
```

In the first 2 scenarios the app never starts (the `listen` method cannot be reached in the first, `port`, `host` and `protocol` are `undefined` in the last).

Perhaps we could a find a better idea to expose the interface. Perhaps we could expose the app as an express `Router`?

JoeChapman:
we could remove the option to not start on initial bootstrap call, we could get extend the defaults with the config passed in to start every time? But I am curious to see what you have in mind re exposing the intercase

danielepolencic:
I wanted to apply a middleware and the only way to do so was to use `use`. Also, I wanted my middleware to be applied before starting the app.

JoeChapman:
ok, so removing the option `start` is off the table

danielepolencic:
I think returning a non started interface could be an option.

JoeChapman:
I think a non-started interface can be useful , esp. because of the scenario you have described re middleware. That is the reason I designed it as such, but I wonder how much people will need that feature. However, I have pushed a commit to the pr that is not particularly elegant, but addresses the issue by conditionally updating the config from the defaults and any new options the user might pass to `start`.

JoeChapman:
Closing as fixed

PR #5, opened 2016-06-21
Update README documentation and optional values

JoeChapman:
- update README
- update how router uses settings, and fallbacks
- hide logger instantiation in test mode

sorry this became a bit larger than I had initially intended, but as I began to update the docs, I realised how little of the config was concrete, and therefore it made little sense to simply update the doc.

### DIFF
--- a/index.js
+++ b/index.js
@@ -3,7 +3,6 @@
 const hof = require('hof');
 const app = require('express')();
 const churchill = require('churchill');
-
 const router = require('./lib/router');
 const serveStatic = require('./lib/serve-static');
 const sessionStore = require('./lib/sessions');
@@ -16,7 +15,8 @@ module.exports = options => {
 
   const load = (config) => {
     config.routes.forEach((route) => {
-      app.use(router(route, config));
+      const routeConfig = Object.assign({route}, config)
+      app.use(router(routeConfig));
     });
   };
 
@@ -31,7 +31,7 @@ module.exports = options => {
         let config = getConfig(options);
       }
       return new Promise((resolve, reject) => {
-        if (config.startOnInitialise === false) {
+        if (config.start === false) {
           return resolve(bootstrap);
         }
         bootstrap.server = require(config.protocol).createServer(app);
@@ -58,20 +58,16 @@ module.exports = options => {
   }
 
   config.routes.forEach(route => {
-    if (!route.fields) {
-      throw new Error('Each route must define a relative path to its fields');
-    }
-    if (!route.views) {
-      throw new Error('Each route must define a relative path to its views');
-    }
     if (!route.steps) {
       throw new Error('Each route must define a set of one or more steps');
     }
-  })
+  });
 
-  const logger = config.logger = require('./lib/logger')(config);
+  if (config.env !== 'test') {
+    const logger = config.logger = require('./lib/logger')(config);
+  }
 
-  if (config.env !== 'ci') {
+  if (config.env !== 'test' && config.env !== 'ci') {
     bootstrap.use(churchill(logger));
   }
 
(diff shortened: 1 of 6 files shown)


index.js
```diff
@@ -16,7 +15,8 @@ module.exports = options => {
 
   const load = (config) => {
     config.routes.forEach((route) => {
-      app.use(router(route, config));
+      config.route = route;
```
danielepolencic: I don't follow. this code is going to overwrite `config.route` for each route. what is it used for?

index.js
```diff
@@ -31,7 +31,7 @@ module.exports = options => {
         let config = getConfig(options);
       }
       return new Promise((resolve, reject) => {
-        if (config.startOnInitialise === false) {
+        if (config.start === false) {
```
danielepolencic: see #4

lib/defaults.js
```diff
@@ -30,7 +30,8 @@ const defaults = {
   },
   session: {
     ttl: process.env.SESSION_TTL || 1800,
-    secret: process.env.SESSION_SECRET || 'changethis'
+    secret: process.env.SESSION_SECRET || 'changethis',
+    key: process.env.SESSION_KEY || 'hod'
```
danielepolencic: I'd rename `key` to `name` since the former isn't officially documented.

index.js
JoeChapman: there is no `config.route` before this. I added this to simplify the arguments passed to router

index.js
JoeChapman: the config is not being generated from the defaults when start is called independently from the initial call to bootstrap

index.js
danielepolencic: if `config.route` is temporary and used only by the current route, perhaps we could write `const routeConfig = Object.assign({route}, config)` and pass that as an argument to the route.

index.js
```diff
@@ -3,20 +3,26 @@
...
 const defaults = require('./lib/defaults');
 
-const getConfig = config => Object.assign({}, defaults, config);
+const getConfig = function (options) {
+  let assigned;
+  [].slice.call(arguments).forEach((arg) => {
+    assigned = Object.assign({}, defaults, arg);
+  })
+  return assigned
+}
```
danielepolencic: `assigned` is overwritten for each argument. When you have `getConfig({a: 1}, {b:2 })` then `assigned` ultimately contains only `Object.assign({}, defaults, {b:2})`. The reference to `{a: 1}` is lost.

An alternative is to use something like:

```js
return Object.assign.apply(null, [{}, defaults].concat(args));
```

JoeChapman: good catch. Would you do the honours and make the change please?

index.js
```diff
@@ -3,20 +3,26 @@
...
 
 module.exports = options => {
 
   const load = (config) => {
     config.routes.forEach((route) => {
-      app.use(router(route, config));
+      const routeConfig = Object.assign({route}, config)
+      app.use(router(routeConfig));
```
danielepolencic: In this case `route` is mutated in place with the values coming from `config`. Is this intentional?

lib/router.js
```diff
@@ -6,43 +6,43 @@ const mixins = hof.mixins;
...
+    fields = require(path.resolve(config.caller, config.fields));
+  } catch (err) {
+    fields = err;
   }
 
-  if (commonFields && fields) {
-    fields = Object.assign({}, commonFields, fields);
+  if ((fields instanceof Error) === false) {
+    fields = Object.assign({}, fields, routeFields);
   }
```
danielepolencic: The `if` expression already evaluates to a `boolean`. It can be replaced with `if(!(fields instanceof Error))`

danielepolencic:
Happy to address the comments if you're ok with it.

JoeChapman:
@danielepolencic please feel free to update code/address comments. Trust you can rebase :)

JoeChapman:
@danielepolencic @easternbloc

danielepolencic:
LGTM 👍

PR #6, opened 2016-06-24
Make the app trust proxy

JoeChapman:
- Trust the proxy (http://expressjs.com/en/guide/behind-proxies.html)


easternbloc:
This is specifically an issue for us because... we use nginx

PR #10, opened 2016-07-01
Errorhandler i18n translate function resolves #7

### LINKED ISSUE #7: Error handler invoked without a translate function
At https://github.com/UKHomeOffice/hof-bootstrap/blob/master/lib/defaults.js#L20 `hof.middleware.error` is invoked with `translate` and `logger`. However, `translate` is `undefined`.
The reason being that `require('hof').i18n` is in fact a function that when invoked returns an object with a translate function. 

Fix:
```
const i18n = require('hof').i18n({
  path: path_to_translations
})

...
...
```

Notes

1): path_to_translations should point to something like... `translations/__lng__/__ns__.json`
Such as:

```
apps/
  |- common/
    |- translations/
      |- en/ 
        |- default.json
```

2) `i18n.translate.bind(i18n)` the translate function must be explicitly bound to the i18n object. When the translate function is invoked without this context, its 'this` object appears to be `undefined`

JoeChapman:
- invoke i18n function with a path to configure translations
- pass i18n translate function to error handler middleware in app
- remove error handler function from user options (./lib/defaults.js)
- update README
- fixed details with logger that was breaking the tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,8 +10,7 @@ You can call the `bootstrap` function with a list of [routes](#routes) and your 
 const bootstrap = require('hof-bootstrap');
 
 bootstrap({
-  views: ...,
-  errorHandler: ...,
+  views: 'optional_path_to_your_views',
   ...,
   routes: [{ ... }, { ... }]
 });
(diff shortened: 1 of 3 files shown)


danielepolencic:
LGTM 👍

PR #11, opened 2016-07-01
Fix the optional paths in the router

JoeChapman:
- modify the test structure to resemble a real service
- include more tests for path options
- make the test directory `test/`
- use fs to test for the rpesence of a file or dir
- throw errors aptly when needed to inform user
- throw an error if the paths are not valid


lib/router.js
```diff
@@ -6,26 +6,46 @@ const mixins = hof.mixins;
...
+  } catch (err) {
+    throw new Error('Cannot find fields');
+  }
 
   try {
-    fields = require(path.resolve(config.caller, config.fields));
+    routeFields = require(paths.routeFields);
   } catch (err) {
-    fields = err;
+    throw new Error('Cannot find route fields');
```
danielepolencic: Since we know the wrong path for route & fields, we can throw a more descriptive error:

```javascript
throw new Error(`Cannot find route fields "${paths.routeFields}"`);
```

same above for fields.

JoeChapman:
^^ @danielepolencic @easternbloc

danielepolencic:
LGTM 👍

JoeChapman:
@easternbloc can you have a look please?

Issue #16, opened 2016-07-07
Use hof-template-partials


joefitter:
fallback views directory should be provided to express application before express-partial-templates middleware. Terms and Cookies should use these templates as a base, also i18n should use common translations compiled from hof-template-partials

PR #18, opened 2016-07-11
Use hof-template-partials for common views

JoeChapman:
resolves https://github.com/UKHomeOffice/hof-bootstrap/issues/16

- install hof-template-partials
- update express-partial-templates

hof-template-partials exports views and translations
This update means we can remove the commonly used views from the service that consumes bootstrap and import the views from hof-template-partials

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,5 +7,9 @@ cache:
   directories:
     - node_modules
 
+before_install:
+  - if [[ `npm -v` != 3* ]]; then npm i -g npm@3; fi
+  - npm cache clean
+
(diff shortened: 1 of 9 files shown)


joefitter:
this LGTM but we should also point to the template-partials for terms/cookies

joefitter:
will submit a PR with these changes on GRO in a sec

JoeChapman:
ok, I'll add that in another pr after this

JoeChapman:
@easternbloc @danielepolencic please take a look at this

PR #26, opened 2016-07-17
Make async tests run and not fail on travis

JoeChapman:
- return test assertion to promise thenable (like calling done)
- set a cookie for each test that makes a get request so as not to fail with a 302
- add cookie middleware to help identify cookie failures (302's)
- add custom route view path to bootstrap functions to override hof-template-partials.views


danielepolencic:
LGTM 👍

PR #27, opened 2016-07-18
Fix asset path

JoeChapman:
- add a slash to local assetPath (`/public`)
- hard-code local assetPath
- make location of assets `assetpath`
- add tests and fixtures for serving assetPath at `/public`
- update test path


joefitter:
LGTM 👍

PR #29, opened 2016-07-19
Remove the `assetPath` option

JoeChapman:
- remove `assetPath` option because there is no way to pass this js option into nginx proxy config
- hard-code mount fs location for express.static to 'public'
- update test and fixture to ensure static is served from `/public`

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,4 +1,5 @@
 node_modules
+test/public
 coverage
 LICENSE
 README.md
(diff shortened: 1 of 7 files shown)


joefitter:
FAILING

JoeChapman:
@joefitter should pass now

JoeChapman:
@joefitter will pass now

joefitter:
It kinda sucks that we have to hard code the public dir but I guess I'm preaching to the choir and you have exhausted all other options, so... LGTM

PR #31, opened 2016-07-19
Use deepTranslate middleware. Resolves #8

### LINKED ISSUE #8: Use deepTranslate middleware
This is applied in `apps/{route}/index.js` before mixins middleware and wizard. A bound instance of the translation function is provided to deepTranslate middleware which is added to `req` object. This is used in template mixins and wizard so no translate method should be additionally supplied to these.

Example - https://github.com/UKHomeOffice/gro/blob/master/apps/gro/index.js#L15-L29

JoeChapman:
- use deepTranslate middleware in route.

### DIFF
--- a/lib/router.js
+++ b/lib/router.js
@@ -51,14 +51,15 @@ module.exports = (config) => {
     path: paths.translations + '/__lng__/__ns__.json'
   });
 
-  router.use(mixins(fields, {
+  router.use(hof.middleware.deepTranslate({
     translate: i18n.translate.bind(i18n)
   }));
 
+  router.use(mixins(fields));
+
   const wizardConfig = {
     controller: config.baseController || hof.controllers.base,
-    templatePath: paths.templates,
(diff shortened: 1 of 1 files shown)


lib/router.js
```diff
@@ -51,6 +51,10 @@ module.exports = (config) => {
     path: paths.translations + '/__lng__/__ns__.json'
   });
 
+  router.use(hof.middleware.deepTranslate({
+    translate: i18n.translate.bind(i18n)
+  }));
+
   router.use(mixins(fields, {
     translate: i18n.translate.bind(i18n)
```
joefitter: the deepTranslate middleware attaches a translate function to `req` which is used in `mixins` - this is overridden if you pass a custom translate function so this should be removed to take advantage

joefitter: This is the same when instantiating wizard too - both give you the option to pass in a custom translate function but fall back to `req.translate`

joefitter:
LGTM

PR #35, opened 2016-08-04
Allow sessionStore to be passed in config

joefitter:
* added .gitignore
* only resolve promise if start === false - we were instantly resolving before.
* fix async issues in tests due to the above change
* apply session middleware with sessionStore and barebones config if ci and sessionStore supplied.

### DIFF
--- a/lib/sessions.js
+++ b/lib/sessions.js
@@ -9,6 +9,21 @@ module.exports = (app, config) => {
 
   const logger = config.logger || console;
 
+  app.use(cookieParser(config.session.secret, {
+    path: '/',
+    httpOnly: true,
+    secure: config.protocol === 'https'
+  }));
+
+  if (config.sessionStore) {
+    return app.use(session({
+      store: config.sessionStore,
+      genid: () => 'fakeId',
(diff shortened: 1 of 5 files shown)


joefitter:
@easternbloc @danielepolencic mind having a 👀  please?

lib/sessions.js
```diff
@@ -9,6 +9,21 @@ module.exports = (app, config) => {
 
   const logger = config.logger || console;
 
+  app.use(cookieParser(config.session.secret, {
+    path: '/',
+    httpOnly: true,
+    secure: config.protocol === 'https'
+  }));
+
+  if (config.env === 'ci' && config.sessionStore) {
```
easternbloc: it'd be nice if we can just inject a sessionStore for stuff that isn't just CI so maybe just remove that check. I get the intention I think it's a nice thing to have

joefitter: YES YOU ARE RIGHT

joefitter:
@easternbloc updated

easternbloc:
👍

PR #43, opened 2016-08-10
Pass name to wizard

joefitter:
* As a router I should pass my name to the wizard if present.
* Wizard assigns a new session key in the format hmpo-wizard={name}
* If name is missing, it increments an integer


joefitter:
@JoeChapman @easternbloc @daniel-ac-martin @danielepolencic quick PR

PR #45, opened 2016-08-12
consolidate variables

danielepolencic:


lib/router.js
```diff
@@ -58,18 +58,12 @@ module.exports = (config) => {
   router.use(mixins(fields));
 
   const wizardConfig = {
+    name,
     controller: config.baseController || hof.controllers.base,
-    templatePath: paths.templates
+    templatePath: paths.templates,
+    params: config.route.params
```
JoeChapman: fails if `params` is `undefined`

PR #46, opened 2016-08-12
Return bootstrap interface object, not a promise

JoeChapman:
### This will be a breaking change

- make interface chaninable by returning bootstrap from interface methods
- remove promise so no more `then`
- apply error middleware in the start method so middleware applied with `bootstrap.use` will be applied before error md.
- move some set up to functions outside of module definition
- update readme documentation
- also, fixed lint error in lib/router

### DIFF
--- a/index.js
+++ b/index.js
@@ -48,7 +48,7 @@ module.exports = options => {
 
   const bootstrap = {
 
-    use: middleware => {
+    use() {
       app.use.apply(app, arguments);
       return bootstrap;
     },
--- a/index.js
+++ b/index.js
@@ -48,7 +48,9 @@ module.exports = options => {
 
   const bootstrap = {
 
-    server: null,
+    app: Object.assign(app, {
+      server: null
+    }),
 
     use: middleware => {
       app.use(middleware);
@@ -62,21 +64,21 @@ module.exports = options => {
 
       const protocol = config.protocol === 'http' ? http : https;
 
-      bootstrap.server = protocol.createServer(app);
+      app.server = protocol.createServer(app);
 
       applyErrorMiddlewares(config, i18n);
 
-      bootstrap.server.listen(config.port, config.host);
+      app.server.listen(config.port, config.host);
       return bootstrap;
     },
 
     stop: (callback) => {
       _.defer(() => {
-        bootstrap.server.close();
+        app.server.close();
       });
 
       if (callback) {
-        callback(bootstrap.server);
+        callback(app.server);
       }
       return bootstrap;
     }
--- a/index.js
+++ b/index.js
@@ -6,6 +6,7 @@ const churchill = require('churchill');
 const path = require('path');
 const http = require('http');
 const https = require('https');
+const _ = require('lodash');
 const router = require('./lib/router');
 const serveStatic = require('./lib/serve-static');
 const sessionStore = require('./lib/sessions');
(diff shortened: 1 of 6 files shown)


index.js
```diff
@@ -1,66 +1,100 @@
...
+    translate: i18n.translate.bind(i18n),
+  }));
+
+  app.use(hof.middleware.errors({
+    translate: i18n.translate.bind(i18n),
+    debug: config.env === 'development'
+  }));
+};
+
+const getI18n = (config) =>
```
joefitter: parens

JoeChapman: glans

index.js
```diff
@@ -1,66 +1,100 @@
...
+  hof.i18n({
+    path: path.resolve(config.caller, config.translations) + '/__lng__/__ns__.json'
```
joefitter: Can we add a check here to test file exists? if not then fallback to hof-template-partials?

index.js
```diff
@@ -1,66 +1,100 @@
...
+      applyErrorMiddlewares(config, getI18n(config));
+
+      bootstrap.server.listen(config.port, config.port);
+      return bootstrap;
     },
 
-    stop: () => {
-      bootstrap.server.close();
+    stop: (callback) => {
+      const interval = setInterval(() => {
```
joefitter: Why is this needed? If we really need to defer can we use `_.defer`

index.js
joefitter: Also - does this need to be in a function if its just called once to set up i18n below?

index.js
```diff
@@ -1,66 +1,100 @@
...
+      if (callback) {
+        callback(bootstrap.server);
+      }
+      return bootstrap;
     }
 
   };
 
   const config = getConfig(options);
+  const i18n = getI18n(config);
```
joefitter: init here?

index.js
JoeChapman: coming in a second PR

index.js
JoeChapman: possibly

index.js
JoeChapman: Can you elaborate?

index.js
JoeChapman: using a function means we can manage variables more easily. Organising groups into functions avoids us relying on outer scopes and stuff.

joefitter: ES6 `const` and `let` are block scoped not function scoped, I may be wrong but this appears to be doing exactly the same as if it were inline, and as it is only called once it is adding complexity (IMO)

index.js
joefitter: Sorry, I just meant the same as above - can't we init inline?

index.js
JoeChapman: IMO, it is easier to manage because we use it in the `start` function; https://github.com/UKHomeOffice/hof-bootstrap/pull/46/files#diff-168726dbe96b3ce427e7fedce31bb0bcR65, otherwise we'd need to pass it into `start`, which becomes very tricky when the user calls `start`

joefitter: But then aren't we init-ing i18n twice?

joefitter: you _could_ or we could simply init it once on calling the bootstrap function an store a reference to it, perhaps on the bootstrap object, to be used in later applied middleware, such as these?

joefitter: Also, do we need to pass config to start? I think that convolutes it somewhat, we are calling bootstrap with config, then calling start with potentially different config. Unless there is a use case I am missing?

JoeChapman: I believe that if a developer wanted to defer start, they might also want to defer passing port, host and/or protocol options

JoeChapman:
Can we rebase these commits please

joefitter:
No Joe, where have you been?

joefitter:
squash and merge 💃

JoeChapman:
squash and merge is what I meant. I have been on holiday. It was nice, thanks for asking

JoeChapman:
Before we merge, is the readme update date?

joefitter:
no Joe, that's your domain

JoeChapman:
Please don't merge this with incorrect docs, it's like merging a commit with failing tests

PR #48, opened 2016-09-07
Enhancement/no more global

joefitter:
This PR scopes the instance of express to the export function so global state is not preserved if multiple instances are required. It also replaces the route `Router` with an instance of express allowing a views dir to be set on the specific route. Express template partials is called per route so partials can also be scoped to journey

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ bootstrap({
 
  * Creates and starts the server listening for connections.
  * `@param {Object}` options
- * `@return {Promise.<bootstrap, Error>}` a promise that returns bootstrap interface object if resolved, Error if rejected.
+ * `@return {Promise.<bootstrap|Error>}` a promise that returns bootstrap interface object if resolved, Error if rejected.
 
 Convenient if starting was deferred during the initial invocation of `hof-bootstrap` with the option and value `start: false` or the server has been stopped. Returns a promise which resolves to the `bootstrap` interface object.
 
@@ -39,7 +39,7 @@ Uses the following settings;
 
  * Closes the server, stops listening for connections
  * `@param {Function}` callback. Useful for testing
- * `@return {Promise.<bootstrap, Error>}` a promise that returns bootstrap interface object if resolved, Error if rejected.
+ * `@return {Promise.<bootstrap|Error>}` a promise that returns bootstrap interface object if resolved, Error if rejected.
 
 ### `use` Function(middleware)
 
--- a/README.md
+++ b/README.md
@@ -24,9 +24,9 @@ bootstrap({
 
  * Creates and starts the server listening for connections.
  * `@param {Object}` options
- * `@return {Object} bootstrap` interface object.
+ * `@return {Promise.<bootstrap, Error>}` a promise that returns bootstrap interface object if resolved, Error if rejected.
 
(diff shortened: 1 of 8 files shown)


joefitter:
@easternbloc @bendiggle @JoeChapman @gxxm @danielepolencic

README.md
```diff
@@ -24,9 +24,9 @@ bootstrap({
 
  * Creates and starts the server listening for connections.
  * `@param {Object}` options
- * `@return {Object} bootstrap` interface object.
+ * `@return {Promise.<bootstrap, Error>}` a promise that returns bootstrap interface object if resolved, Error if rejected.
```
JoeChapman: Please can you use a pipe character (`|`) to delimit optional arguments

README.md
```diff
@@ -39,7 +39,7 @@ Uses the following settings;
 
  * Closes the server, stops listening for connections
  * `@param {Function}` callback. Useful for testing
- * `@return {Object} bootstrap` interface object.
+ * `@return {Promise.<bootstrap, Error>}` a promise that returns bootstrap interface object if resolved, Error if rejected.
```
JoeChapman: Please can you use a pipe character (`|`) to delimit optional arguments

README.md
joefitter: I don't follow, promise notation isn't very well documented - I went with how Medium do it - https://github.com/jsdoc3/jsdoc/issues/509#issuecomment-29186993. Could you give an example if we do it differently?

README.md
joefitter: as above

README.md
joefitter: do you mean in between `bootstrap` and `Error`?

JoeChapman: That conversation never really concludes, but examples include syntax with a pipe, which I prefer to comma because it's an either or scenario and comma used in other contexts express multiple arguments not binary. `@return {Promise.<bootstrap|Error>}`

Does that make sense, how do you feel about this?

joefitter: sure makes sense - thanks for clarifying!

lib/router.js
```diff
@@ -52,15 +55,18 @@ module.exports = (config) => {
...
     translate: i18n.translate.bind(i18n)
   }));
 
-  router.use(mixins(fields));
+  app.use(mixins(fields));
 
   const wizardConfig = {
-    controller: config.baseController || hof.controllers.base,
-    templatePath: paths.templates
+    controller: config.baseController || hof.controllers.base
```
JoeChapman: As we set baseController to hof.controller.base in defaults.js, this could be `controller: config.baseController`

lib/router.js
```diff
@@ -71,8 +77,8 @@ module.exports = (config) => {
     wizardConfig.params = config.route.params;
   }
 
-  router.use(baseUrl, wizard(config.route.steps, fields, wizardConfig));
```
JoeChapman: is baseUrl no longer used?

joefitter: not here, its used in index.js so we are able to scope per route

joefitter: here https://github.com/UKHomeOffice/hof-bootstrap/pull/48/files#diff-168726dbe96b3ce427e7fedce31bb0bcR27

lib/router.js
joefitter: 👍

PR #58, opened 2016-10-07
trust all proxies to allow for many hops

JoeChapman:


easternbloc:
cherry pick the commit from master.

PR #73, opened 2016-11-22
Update healthcheck endpoint to also report uptime.

easternbloc:
This middleware also gives us hook in the future to check other layers for a deeper healthcheck.

### DIFF
--- a/index.js
+++ b/index.js
@@ -98,7 +98,7 @@ module.exports = options => {
   };
 
   // shallow health check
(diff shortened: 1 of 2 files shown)


JoeChapman:
Have you considered: https://github.com/UKHomeOfficeForms/hof-bootstrap/issues/69?

easternbloc:
I have I just wanted a more useful shallow one atm but this has the facility to handle that once we add it

PR #75, opened 2016-11-23
Update documentation

JoeChapman:
- include quick find links
- update badge for UKHomeOfficeForms
- include steps to migrate from HOF
- include reference for newbies
- include resources links
- add tests for multiple sub apps/routes


README.md
```diff
@@ -1,32 +1,133 @@
...
+`cd` into your project and install HOF-bootstrap.
+```
+$ npm install --save hof-bootstrap
+```
+
+### From HOF to HOF-bootstrap
+If you are familiar with HOF, the process of moving to HOF-bootstrap should be straightforward. The entry point of your app, will contain the boilerplate code that configures the apps' settings, routes, middleware, session storage and probably starts the server listening for connections.
+You can safely remove the boilerplate code and replace it with the bootstrap function. The following example invokes `bootstrap` with [routes](#routes). The custom configuration is used to tell HOF-bootstrap where to look for resources such as the views or controllers, or what session store to use.
+
+HOF-bootstrap includes all the same dependencies that HOF contains, so you can remove the HOF dependency from `package.json` and `npm uninstall --save hof`.
```
joefitter: also hmpo-frontend-toolkit is not included in bootstrap - this is as it is only css and client side JS, and not needed in bootstrap. This needs to be installed in the application separately if required

README.md
```diff
@@ -124,27 +201,41 @@ Not all route settings are mandatory, you can create and launch a service with j
...
+## Resources
+
+### HOF documentation
+https://github.com/UKHomeOfficeForms/HOF/blob/master/documentation/index.md
+
+## Services that are built with HOF-bootstrap
+- https://github.com/UKHomeOffice/gro
+- https://github.com/UKHomeOffice/rotm
+- https://github.com/UKHomeOffice/end-tenancy
+- https://github.com/UKHomeOffice/UKVI-Complaints
```
joefitter: firearms

README.md
JoeChapman: Perhaps I should update the engine value in the package to a version that requires npm 3?

JoeChapman: I can specify the version of npm a la: https://docs.npmjs.com/files/package.json#engines

joefitter: yes that should do it :)

JoeChapman: done

README.md
JoeChapman: nah

joefitter: lol

PR #79, opened 2016-11-23
feature - added example form

joefitter:
Added boilerplate code and a couple of simple steps to demonstrate how to create a form using bootstrap. Added codecept tests

TODO:
* add more hof features
* add a multi-journey form example


example/apps/example-app/index.js
```diff
@@ -0,0 +1,54 @@
...
+      ],
+      next: '/confirm',
+      locals: {
+        section: 'enquiry-details'
+      }
+    },
+    '/confirm': {
+      controller: controllers.confirm,
+      fieldsConfig: require('./fields'),
+      emailConfig: require('../../config').email,
```
JoeChapman: It would be nice, if this didn't need to be explicitly defined, if the confirm step would look for the email config.

example/index.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
+
+const bootstrap = require('hof-bootstrap');
+
+bootstrap({
+  views: false,
+  fields: false,
```
JoeChapman: I'm going to fix this tomorrow so we don't need to pass `false`

example/apps/example-app/index.js
joefitter: yes - good point - it would actually be better to include a base config in hof-controllers, and then we can extend it if we need to

example/package.json
```diff
@@ -0,0 +1,37 @@
...
+    "watch:scss": "nodemon -e scss -x 'npm run sass'",
+    "watch:js": "nodemon --watch assets/js -e js -x 'npm run browserify'",
+    "watch:translations": "nodemon --watch apps/**/translations/src -x 'npm run hof-transpile'",
+    "hof-transpile": "hof-transpiler ./apps/**/translations/src -w --shared ./node_modules/hof-template-partials/translations",
+    "sass": "node-sass ./assets/scss/app.scss ./public/css/app.css --include-path ./node_modules",
+    "browserify": "browserify ./assets/js/index.js > ./public/js/bundle.js",
+    "copy:images": "cp -r ./assets/images ./public/",
+    "create:public": "mkdir -p ./public/js ./public/css ./public/images",
+    "postinstall": "npm run create:public; npm run sass; npm run browserify; npm run copy:images; npm run hof-transpile",
+    "codecept": "codeceptjs run ./node_modules/so-acceptance --steps",
```
JoeChapman: I think this should be called `acceptance`

example/index.js
joefitter: nice one - and while you're at it, it shouldn't throw an error if there isn't a views directory

example/package.json
joefitter: wateva

JoeChapman:
Awesome

example/apps/example-app/index.js
JoeChapman: yep

example/index.js
JoeChapman: yes, yes, yes

Issue #68, opened 2016-11-08
Remove root level views and fields 


JoeChapman:
Do not make bootstrap expect views, fields and translations at the root level of a project.
At the moment, bootstrap will throw an error if they don't exist. However, it would make sense to throw an error only if no views, or fields, or translations could be found at the root level or the route level

PR #85, opened 2016-12-02
Make views and fields use defaults properly

### LINKED ISSUE #52: Dont throw error if route views are missing
This is a feature. Views do not need to be provided if they exist in common or hof-template-partials.

JoeChapman:
- don't need to add false to base level fields and views options
- uses default fields and views paths as last resort.
- does not throw if the default paths exist
- move getPaths, getViews, and getFields into helper class as statics
- update readme

EDIT:
- remove default fields and views values from defaults.js
- throw if no fields specified

Resolves #52 and #68

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,10 +156,10 @@ HOF-bootstrap accepts the following options so a developer can customise element
 
 ### Options
 
-- `views`: Location of the base views relative to the root of your project. Defaults to 'views'. Set `views` to `false` if not present.
+- `views`: Location of the base views relative to the root of your project. Defaults to `__dirname + '/views'`.
+- `fields`: Location of the common fields relative to the root of your project. Defaults to `__dirname + '/fields'`.
+- `translations`: Location of the common translations relative to the root of your project. Defaults to `./translations`.
 - `middleware`: An optional array of middleware functions to add to the application middleware pipeline.
-- `fields`: Location of the common fields relative to the root of your project. Defaults to 'fields'. Set `fields` to `false` if not present.
-- `translations`: Location of the common translations relative to the root of your project. Defaults to 'translations'.
 - `baseController`: The base controller for all routes and steps. Defaults to [HOF-controllers.base](https://github.com/UKHomeOfficeForms/hof-controllers/blob/master/lib/base-controller.js).
 - `viewEngine`: Name of the express viewEngine. Defaults to 'html'.
 - `start`: Start the server listening when the bootstrap function is called. Defaults to `true`.
@@ -223,12 +223,28 @@ const myRoutes = [{
 [Read more about steps and fields](https://github.com/UKHomeOfficeForms/HOF/blob/master/documentation/index.md)
 
 #### Options
-- `name`: If provided, is used to locate views, fields and translations for a form journey.
+- `name`: Passed to the form Wizard.
 - `baseUrl`: Base url from which all steps are relative. Defaults to `/`. If provided will be used to locate views, fields and translations for a form journey.
-- `fields`: Location of the routes' fields, relative to the root of your project. Defaults to `fields`.
-- `views`: Location of the routes' views relative to the root of your project. Defaults to `views`.
+- `fields`: Location of the routes' fields, relative to the root of your project. Defaults to `__dirname + '/fields'`.
+- `views`: Location of the routes' views relative to the root of your project. Defaults to `__dirname + '/views'`.
 
-**NOTE**: `fields` defined in a `route` determine the name of the directory or path, relative to the root, where the `fields` module is located. `fields` defined in a step, are a list of the name of each field you want to use in the step.
+**NOTE**: The `fields` defined in a `route` is the path to the folder, relative to the root, where the `fields` are is located.
+The `fields` defined in a `step`, are a list of the name of each field you want to use in the step.
+
+For example:
+```
+[{
+  fields: '../../fields',
+  steps: {
+    '/one': {
+      fields: [
+        'name_of_field_one',
+        'name_of_field_two'
+      ]
+    }
+  }
+}];
+```
 
 ## Resources
 
--- a/README.md
+++ b/README.md
@@ -156,8 +156,15 @@ HOF-bootstrap accepts the following options so a developer can customise element
 
 ### Options
 
-- `views`: Location of the base views relative to the root of your project. Defaults to `__dirname + '/views'`.
+- `views`: Location of the base views relative to the root of your project.
+  - Will not error if it can't find any views in the root of your project.
+  - Looks for views - optionally set in your [route options](#route-options).
+  - Always loads views from [hof-template-partials.views](https://github.com/UKHomeOfficeForms/hof-template-partials/tree/master/views).
+  - Views extend like so, where route-views are the most specific: `hot-template-partials -> your-common-views -> your-route-views`
+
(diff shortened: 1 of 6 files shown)


lib/helpers.js
```diff
@@ -0,0 +1,69 @@
...
+        throw new Error(`Cannot find views at ${pathViews.base}`);
+      }
+    }
+
+    if (pathViews.route) {
+      try {
+        fs.accessSync(pathViews.route, fs.F_OK);
+        views.unshift(pathViews.route);
+      } catch (err) {
+        throw new Error(`Cannot find route views at ${pathViews.route}`);
```
JoeChapman: This only throws if someone has passed a path to views, which is incorrect

lib/helpers.js
```diff
@@ -0,0 +1,69 @@
...
+
+    return Object.assign({}, fields, routeFields);
+  }
+
+  static getViews(pathViews) {
+    let views = [];
+
+    if (pathViews.base) {
+      try {
+        fs.accessSync(pathViews.base, fs.F_OK);
```
joefitter: you can call `fs.accessSync(pathViews.base)` without passing `fs.F_OK` as this is the default option if you omit the argument

joefitter: no biggie, no harm in being verbose I guess, just saving chars!

JoeChapman: ok I see, cheers

lib/helpers.js
```diff
@@ -0,0 +1,69 @@
...
+        views.unshift(pathViews.route);
+      } catch (err) {
+        throw new Error(`Cannot find route views at ${pathViews.route}`);
+      }
+    }
+
+    return views;
+  }
+
+};
```
JoeChapman: Yes, that is something I might investigate. However, it will introduce breaking changes, that this doesn't. I'd like to treat the two concepts (this and `require` instead of paths) as two different PR's, alright?

joefitter: alrighty then!

JoeChapman:
Whoop! @easternbloc, @bendiggle, @danielepolencic @sulthan-ahmed

README.md
```diff
@@ -156,10 +156,10 @@ HOF-bootstrap accepts the following options so a developer can customise element
 
 ### Options
 
-- `views`: Location of the base views relative to the root of your project. Defaults to 'views'. Set `views` to `false` if not present.
+- `views`: Location of the base views relative to the root of your project. Defaults to `__dirname + '/views'`.
+- `fields`: Location of the common fields relative to the root of your project. Defaults to `__dirname + '/fields'`.
+- `translations`: Location of the common translations relative to the root of your project. Defaults to `./translations`.
```
sulthan-ahmed: I noticed you removed `Set `views` to `false` if not present.` So what do you do now if your views are not present?

sulthan-ahmed: I noticed you removed ```Set `views` to `false` if not present.```. So what do you do now if your views are not present?

README.md
```diff
@@ -223,12 +223,28 @@ const myRoutes = [{
 [Read more about steps and fields](https://github.com/UKHomeOfficeForms/HOF/blob/master/documentation/index.md)
 
 #### Options
-- `name`: If provided, is used to locate views, fields and translations for a form journey.
+- `name`: Passed to the form Wizard.
```
sulthan-ahmed: What is the name?

README.md
```diff
@@ -223,12 +223,28 @@ const myRoutes = [{
...
 - `baseUrl`: Base url from which all steps are relative. Defaults to `/`. If provided will be used to locate views, fields and translations for a form journey.
-- `fields`: Location of the routes' fields, relative to the root of your project. Defaults to `fields`.
-- `views`: Location of the routes' views relative to the root of your project. Defaults to `views`.
+- `fields`: Location of the routes' fields, relative to the root of your project. Defaults to `__dirname + '/fields'`.
+- `views`: Location of the routes' views relative to the root of your project. Defaults to `__dirname + '/views'`.
 
-**NOTE**: `fields` defined in a `route` determine the name of the directory or path, relative to the root, where the `fields` module is located. `fields` defined in a step, are a list of the name of each field you want to use in the step.
+**NOTE**: The `fields` defined in a `route` is the path to the folder, relative to the root, where the `fields` are is located.
```
sulthan-ahmed: is this a typo `are is`, do you mean to say `are`

README.md
```diff
@@ -223,12 +223,28 @@ const myRoutes = [{
...
+  steps: {
+    '/one': {
+      fields: [
+        'name_of_field_one',
+        'name_of_field_two'
+      ]
+    }
+  }
+}];
+```
```
sulthan-ahmed: I like examples 👍

README.md
JoeChapman: It will use the views in `hof-template-partials`, which I can document

README.md
JoeChapman: Whatever you want it to be

README.md
JoeChapman: yes

README.md
joefitter: the views in hof-template-partials are always used implicitly, this views option on the root level is to provide a directory of common views shared between different routes in the same app.

joefitter: without root views extension is `hof-template-partials -> your-route-views`, if you give it a common views folder it will be `hot-template-partials->your-common-views->your-route-views`

sulthan-ahmed: Ahh so this PR resolves the issue I had last week where I didn't have a views folder.  Sweet Yes this is helpful to know and documented thank you both @JoeChapman and @joefitter

JoeChapman:
@joefitter @sulthan-ahmed @easternbloc I am updating this; removing the detault `fields` and `views` values.
Should bootstrap throw an error if no `fields` are set or just a console warning?

joefitter:
it is possible that a route would have no fields, but this would only work if common fields are set on the base config object. If you want to be clever you could throw an error if no fields are found for the route and no common fields have been set, but easier option could be just allow no fields to be set

joefitter:
views shouldn't warn or throw an error IMO as if omitted it will always use hof-template-partials

JoeChapman:
thanks @joefitter. Views should never throw, and fields will throw if none set for either route or base.

joefitter:
LGTM!

README.md
```diff
@@ -156,10 +156,17 @@ HOF-bootstrap accepts the following options so a developer can customise element
 
 ### Options
 
-- `views`: Location of the base views relative to the root of your project. Defaults to 'views'. Set `views` to `false` if not present.
+- `views`: Location of the base views relative to the root of your project.
+  - Will not error if it can't find any views in the root of your project.
+  - Looks for views - optionally set in your [route options](#route-options).
+  - Always loads views from [hof-template-partials.views](https://github.com/UKHomeOfficeForms/hof-template-partials/tree/master/views).
```
sulthan-ahmed: always? I'm confused.  Won't it look at the hof-partials by default unless you create your own views

JoeChapman: nope, always

joefitter: ```hof-template-partials -> route-views```
or
```hof-template-partials -> common-views -> route-views```

JoeChapman: [hof-template-partials.views] are always loaded.
So are base views and route views, if they a path to them if specified in the options, and the path is valid

PR #101, opened 2017-01-04
bugfix - fix broken paths to views and translations

joefitter:
Since bootstrap v7 there have been a number of issues around implied paths. Previously it was assumed that `./views`, `./translations` and `./fields` would be relative to the path of the route. The previous change removed this and required them to be explicitly set. This commit allows them to be explicitly set, but if not it will attempt to locate them in the route directory as before. This actually removes the need for the previous major bump

* updated to use name if fields translations and views are not explicitly set
* removed duplicate shared views path

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -3,28 +3,35 @@
 const fs = require('fs');
 const path = require('path');
 
-module.exports = class Helpers {
-
-  static getRouteFields(config) {
-    let routeFields = config.route.fields && path.resolve(config.caller, config.route.fields);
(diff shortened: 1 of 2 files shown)


lennym:
It'd be cool to have an idea of what the bug being fixed was, but otherwise the code LGTM. 🚀

joefitter:
Yeah - I am not known for my overly descriptive commit messages 😬 - I'll add to the description now

lennym:
Looks legit then.

PR #107, opened 2017-01-12
Add basic Content Security Policy (CSP) middleware

JoeChapman:
- CSP middlware defines a whitelist of origins assets can be loaded from.
- The script directive includes `"'unsafe-inline'"` due to the inline scripts
  inherited from govuk_templates (https://github.com/alphagov/govuk_template/issues/258).

- Remove jscs because it dupes eslint

Resolves: https://github.com/UKHomeOfficeForms/hof-bootstrap/issues/105

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,9 +17,8 @@
     "url": "https://github.com/UKHomeOffice/hof-bootstrap/issues"
   },
   "scripts": {
-    "test": "npm run unit && npm run style && npm run lint",
+    "test": "npm run unit && npm run lint",
     "unit": "NODE_ENV=test istanbul cover _mocha",
-    "style": "jscs index.js **/*.js test --config=./.jscsrc",
     "lint": "eslint .",
     "snyk": "snyk test",
     "snyk:dev": "snyk test --dev",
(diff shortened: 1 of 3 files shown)


JoeChapman:
I know, but it failed the jscs check which duplicated a check run by eslint that had passed. That paired with the fact that we had spoken about removing jscs because we shouldn't need both jscs and eslint.

joefitter:
fair

JoeChapman:
@easternbloc @lennym @sulthan-ahmed @gxxm please review?

package.json
```diff
@@ -17,9 +17,7 @@
     "url": "https://github.com/UKHomeOffice/hof-bootstrap/issues"
   },
   "scripts": {
-    "test": "npm run unit && npm run style && npm run lint",
-    "unit": "NODE_ENV=test istanbul cover _mocha",
-    "style": "jscs index.js **/*.js test --config=./.jscsrc",
+    "test": "NODE_ENV=test istanbul cover _mocha",
```
lennym: I'd probably still want `npm test` to run linting.

JoeChapman:
@lennym gets what @lennym wants

lennym:
Now you've half removed jscs.

Need to remove `npm run style` from test script.

JoeChapman:
I don't know what you are talking about @lennym

PR #108, opened 2017-01-13
Throw error if session.secret is not set in production

lennym:
This provides some bullet-proof socks for people accidentally starting insecure apps in production environments.

### DIFF
--- a/lib/sessions.js
+++ b/lib/sessions.js
@@ -9,6 +9,10 @@ module.exports = (app, config) => {
 
   const logger = config.logger || console;
 
+  if (config.env === 'production' && config.session.secret === 'changethis') {
+    throw new Error('Session secret must be set to a unique random string for production environments');
+  }
+
   app.use(cookieParser(config.session.secret, {
     path: '/',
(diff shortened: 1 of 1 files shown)


lennym:
Spotted we were hacking sessions in acceptance tests. Became mildly concerned how easy that is. Did this.

Only possible issue: do we always set `NODE_ENV=production`?

JoeChapman:
> Only possible issue: do we always set NODE_ENV=production?

In production? yes, *we* do

lennym:
That's cool. I was also interested in possible `prod` vs `production` ambiguity.

JoeChapman:
Ah, I see.
We have and intend to always use `production` not `prod`. 
I suppose we could look to template more of the config to mitigate ambiguities such as these.

lennym:
Maybe in the medium/long term that would be a good idea, but let's not overthink it for now. This was just meant to be an easy win.

_(I will be interested to see if it brings anything down though ;-))_

JoeChapman:
_evil laugh_

PR #110, opened 2017-01-16
changelog for upcoming version (v8.1.0)

JoeChapman:
Changes listed

* Add basic Content Security Policy (CSP) middleware
* Remove JSCS in favour of only Eslint
* Throw error if session.secret is not set in production
* Use snyk on master branch in travis CI


lennym:
Should we limit changelog only to things that affect usage?

The snyk and jscs things are worthwhile from a dev perspective, but not really of any value to end users/implementers, who are the target audience for a changelog.

JoeChapman:
Yeah @lennym, actually I think that is a good idea

PR #112, opened 2017-01-19
Remove environment variable based root directory detection

lennym:
Having the project root directory be dependent on the NODE_ENV environment variable works fine when testing hof-bootstrap itself, but causes unexpected behaviour when testing apps which are dependent on hof-bootstrap.

Instead make the project root explicitly configurable, with a default of the current working directory.

Add a convenience method to the exported bootstrap module to allow for modifying global configuration. This allows for the default project root to be set in the integration tests without needing to be set in the constructor in each instantiation.

*Might* be a breaking change - `caller` config point isn't documented, so probably isn't used anywhere. Certainly isn't in any of our current workstreams.

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,6 +156,9 @@ HOF-bootstrap accepts the following options so a developer can customise element
 
 ### Options
 
+- `root`: The path to the project root directory.
+  - App paths will be relative to this directory.
+  - Defaults to the current working directory.
 - `views`: Location of the common views relative to the root of your project.
   - Will not error if it can't find any views in the root of your project.
   - Looks for views - optionally set in your [route options](#route-options).
--- a/README.md
+++ b/README.md
@@ -15,6 +15,7 @@ HOF-bootstrap is a wrapper function for HOF. It takes the hard work out of setti
   - [server](#server)
 - [Custom configuration](#custom-configuration)
   - [Options](#options)
+  - [Global configuration](#global-configuration)
   - [Environment variables](#environment-variables)
 - [Routes](#routes)
 - [Resources](#resources)
@@ -171,7 +172,8 @@ HOF-bootstrap accepts the following options so a developer can customise element
 - `middleware`: An optional array of middleware functions to add to the application middleware pipeline.
 - `baseController`: The base controller for all routes and steps. Defaults to [HOF-controllers.base](https://github.com/UKHomeOfficeForms/hof-controllers/blob/master/lib/base-controller.js).
 - `appConfig`: Allows you to attach a configuration object to each controllers' options argument. Useful if you need to access properties of your applications config settings in other parts of your code base, e.g:
-```
+
+```javascript
 ...
 constructor(options) {
   this.emailSettings = options.appConfig.emailSettings;
@@ -194,6 +196,16 @@ constructor(options) {
 - `session.ttl`: The session timeout in milliseconds. Defaults to `1800` (ms).
 - `session.secret`: The session secret. Set this to something unique.
 - `session.name`: The session name. Set this to something unique.
+
+### Global configuration
+
+Any of the configuration parameters can be set globally for all bootstrap instances by calling `bootstrap.configure`. This is mostly relevant for running unit tests which will create a large number of bootstrap instances.
+
+```javascript
+const bootstrap = require('hof-bootstrap');
+// set the project root to a custom location for all bootstrap instances
+boostrap.configure('root', path.resolve(__dirname, '..'));
+```
 
 ### Environent variables
 
--- a/index.js
+++ b/index.js
@@ -156,7 +156,7 @@ function bootstrap(options) {
 bootstrap.configure = function configure(key, val) {
   if (arguments.length === 2 && typeof key === 'string') {
     customConfig[key] = val;
-  } else {
+  } else if (typeof key === 'object') {
     Object.assign(customConfig, key);
   }
 };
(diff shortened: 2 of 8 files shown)


index.js
```diff
@@ -121,32 +123,42 @@ module.exports = options => {
...
-  return bootstrap;
+  return instance;
+
+}
 
+bootstrap.configure = function configure(key, val) {
+  if (arguments.length === 2 && typeof key === 'string') {
+    customConfig[key] = val;
+  } else {
+    Object.assign(customConfig, key);
```
JoeChapman: what if `key` is a string?

lib/defaults.js
```diff
@@ -1,21 +1,14 @@
...
 const defaults = {
+  root: process.cwd(),
   translations: 'translations',
-  caller: caller,
   start: true,
   getCookies: true,
   getTerms: true,
   viewEngine: 'html',
-  baseController: controllers.base,
+  baseController: require('hof-controllers').base,
```
JoeChapman: why move this `require`?

lennym: Because it's a single-use variable. There didn't seem any value in the extra line of code as I don't think removing the variable impacted readability.

index.js
lennym: Then the config ends up a but weird, but nothing bad happens. Will add a `typeof` check though.

index.js
```diff
@@ -121,32 +123,42 @@ module.exports = options => {
...
-    bootstrap.start(config);
+    instance.start(config);
   }
 
-  return bootstrap;
+  return instance;
+
+}
 
+bootstrap.configure = function configure(key, val) {
```
JoeChapman: Can you document this?

lennym: Yes.

JoeChapman:
`caller` is not used by any implementation (disclaimer - afaik)

lennym:
I'd have been surprised if it were, it's a bit weird.

README.md
```diff
@@ -195,6 +196,16 @@ constructor(options) {
...
+### Global Configuration
+
+Any of the configuration parameters can be set globally for all bootstrap instances by calling `bootstrap.configure`. This is mostly relevant for running unit tests which will create a large number of bootstrap instances.
+
+```javascript
+const bootstrap = require('hof-bootstrap');
+// set the project root to a custom location for all bootstrap instances
+boostrap.configure('root', path.resolve(__dirname, '..'));
+```
+
```
JoeChapman: Thanks @lennym please add a link to Global Configuration from the menu at the top

lennym: Done.

index.js
```diff
@@ -15,9 +15,11 @@ const defaults = require('./lib/defaults');
 const logger = require('./lib/logger');
 const helmet = require('helmet');
 
+const customConfig = {};
+
 const getConfig = function getConfig() {
   const args = [].slice.call(arguments);
-  return Object.assign.apply(null, [{}, defaults].concat(args));
+  return Object.assign.apply(null, [{}, defaults, customConfig].concat(args));
```
joefitter: Couldn't this simply be:

```js
Object.assign({}, defaults, customConfig, args)
```

lennym: No, `args` is an array of config objects.

lennym: I'd personally remove the case where this function can be called with more than one argument (namely when passing additional config to `start`), and hence be able to clean up this code, but it's out of scope of this PR.

lib/defaults.js
joefitter: this is fine while using `require`, but when/if es6 `import` ever ships, all imports are hoisted to the top of the file.

lennym: I'll write code for if/when `import` ever ships, if/when `import` ever ships.

index.js
joefitter: bootstrap _could_ be an es6 `class`, this could be a `static` method... discuss!

joefitter: obvs it would then need to be instantiated with `new` in apps

JoeChapman: two boostrap instances per app?

lennym: @joefitter What would the benefit of that be?

Sounds like "hacker news driven development" to me.

index.js
joefitter: ah of course it is, and yes, out of scope, this does seem to be needlessly complex though

lennym:
Yeah, `caller` immediately offended me too.

lib/defaults.js
joefitter: lol

index.js
joefitter: hahaha get with the times ;)

lib/defaults.js
joefitter: fair, personally I like the pattern of keeping all imports at the top of the file so you can immediately see all dependencies, useful when mocking etc. Again its a preference thing, and I totally get your point that it adds an extra var and LOC

JoeChapman:
be offended

index.js
lennym: I think it's needfully complex, but the need itself is needless.

index.js
lennym: I just came from there.

lib/defaults.js
lennym: I largely agree, except for short files, and in particular where no function calls are made. As soon as you're inside a function scope then you should be require-ing anything because blocking, but for something like this where everything is executed immediately my official position is ¯\_(ツ)_/¯

PR #113, opened 2017-01-27
Major version update of hof-form-wizard

lennym:
Includes the controller refactor from the latest version of hof-form-wizard. This removes the need to include hof-controllers as a direct dependency since the base controller is now exposed directly from hof-form-wizard.

This now allows dynamic controller config to be used in hof-bootstrap apps.

🎉🎉🎉🎉🎉


lennym:
Actually, I should probably npm link this into a project and run acceptance tests against it first... please hold.

PR #114, opened 2017-01-27
Allow CSP to be disabled

lennym:
CSP prevents inline script injection, but running an automated browser for acceptance tests requires script injection. Allow CSP to be disabled so that we can run acceptance tests against apps.


lennym:
Fix for failing linting is in #113

PR #116, opened 2017-02-03
Allow Content Security Policy directives to be extended with config option

JoeChapman:
The default directives can be extended when required rather than adding more defaults to core that would likely be implementation oriented. Such as adding www.google.com... as a an allowed src for the analytics code

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,6 +180,10 @@ constructor(options) {
 }
 ...
 ```
+- `csp` (We use [Helmetjs/csp](https://github.com/helmetjs/csp) middleware)
+  - `true` by default, enables Content Security Policy middleware.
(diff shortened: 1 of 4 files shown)


lennym:
I feel like having an additional option to `config.csp` of `config.extraCSP` is kinda nasty and redundant.

I'd prefer something like:

* `config.csp` is falsy => csp disabled
* `config.csp` is *true* =>  csp enabled with default settings
* `config.csp` is an object of the appropriate format => csp is enabled with custom extended settings.

README.md
```diff
@@ -180,7 +180,7 @@ constructor(options) {
 }
 ...
 ```
-
+- `extraCSP`: Extend the default Content Security Policy directives with your own.
```
lennym: Update this.

PR #120, opened 2017-02-09
Add google as a source to directives for imag and script when `gaTagI…

JoeChapman:
…d` is set

When `gaTagId` is set then the google analytics script is included in the  view template, which in turn loads it's own script and image from google. This change also adds google.... to the CSP directives meaning those scripts can be loaded and google analytics will work.

shizzle to @lennym for the suggestion

### DIFF
--- a/index.js
+++ b/index.js
@@ -45,16 +45,27 @@ const applyErrorMiddlewares = (app, config, i18n) => {
   }));
 };
 
-const getContentSecurityPolicy = (csp) => {
+const getContentSecurityPolicy = config => {
+  let csp = config.csp;
   let directives = {
     /* eslint-disable quotes */
     defaultSrc: ["'none'"],
     styleSrc: ["'self'"],
     imgSrc: ["'self'"],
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -45,13 +45,20 @@ const applyErrorMiddlewares = (app, config, i18n) => {
...
 
-const getContentSecurityPolicy = (csp) => {
+const getContentSecurityPolicy = (config) => {
+  let csp = config.csp;
+  if (config.gaTagId) {
+    csp = {
+      scriptSrc: ['http://www.google-analytics.com/analytics.js'],
+      imgSrc: ['http://www.google-analytics.com/collect']
+    };
+  }
```
lennym: I don't think this is correct. It's blowing away any user-provided config for csp if `gaTagId` is set.

In particular, if a user has provided custom csp *and* a `gaTagId` then both should be respected, and the google analytics sources should be added to the user-provided csp.

JoeChapman: good point

PR #122, opened 2017-02-10
Make GA CSP directive urls broader to allow google to include js and images

JoeChapman:
Discovered in the real world that google analytics script loads image from unexpected path.
By allowing google to load scripts and images we remove the overhead of regularly updating bootstrap or defining custom csp at implementation, so...

- remove protocol from google url
- remove path from google url
- include note in docs

### DIFF
--- a/test/integration/index.js
+++ b/test/integration/index.js
@@ -541,8 +541,8 @@ describe('bootstrap()', () => {
         .set('Cookie', ['myCookie=1234'])
         .expect((res) => {
           const csp = getHeaders(res, 'content-security-policy');
-          csp['img-src'].should.include('http://www.google-analytics.com/collect');
-          csp['script-src'].should.include('http://www.google-analytics.com/analytics.js');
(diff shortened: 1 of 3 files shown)


test/integration/index.js
```diff
@@ -566,8 +566,8 @@ describe('bootstrap()', () => {
         .set('Cookie', ['myCookie=1234'])
         .expect((res) => {
           const csp = getHeaders(res, 'content-security-policy');
-          csp['img-src'].should.include('http://www.google-analytics.com/collect', 'bar', 'self', 'data:');
-          csp['script-src'].should.include('http://www.google-analytics.com/analytics.js', 'poo', 'self');
+          csp['img-src'].should.include('www.google-analytics.com', 'bar', 'self', 'data:');
+          csp['script-src'].should.include('www.google-analytics.com', 'poo', 'self');
```
lennym: This test is probably not testing what you think it is. There's a typo `foo` => `poo` which *should* fail the test, but currently is not.

PR #125, opened 2017-02-15
Feature - Behaviours, Hooks and Components

joefitter:
* Removed hof-controlers
* Bumped hof-form-wizard to 3.0.0
* Bumped hof-template-mixins to 3.1.0
* Removed default controller
* Tweaked tests to use behaviours rather than custom controllers
* Added hof-component-date and hof-behaviour-confirm-page to example app
* Added date component example to example app
* Shaved many a yak


example/apps/example-app/translations/src/en/fields.json
```diff
@@ -1,20 +1,28 @@
...
+    "legend": "Enter a date in the past"
+  },
+  "date-example-day": {
+    "label": "Day"
+  },
+  "date-example-month": {
+    "label": "Month"
+  },
+  "date-example-year": {
+    "label": "Year"
```
joefitter: These labels should be set as defaults in the component - I'll add an issue

joefitter:
@lennym would like your 👌 when you get time

example/apps/example-app/index.js
```diff
@@ -1,49 +1,53 @@
...
-const controllers = require('hof-controllers');
+const SummaryPageBehaviour = require('hof-behaviour-summary-page');
 
 module.exports = {
   name: 'example-app',
   params: '/:action?',
   steps: {
     '/': {
-      controller: controllers.start,
+      behaviours: require('./behaviours/clear-session'),
```
lennym: I think it'd be good if we used https://github.com/UKHomeOfficeForms/hof-form-wizard/pull/11 in the example app instead of something custom to achieve the same goal.

example/apps/example-app/index.js
```diff
@@ -1,49 +1,53 @@
...
-      emailConfig: require('../../config').email,
+      behaviours: SummaryPageBehaviour,
+      sections: {
+        'personal-details': [
+          'your-name',
+          'email-address',
+          'phone-number'
+        ],
+        'enquiry-details': [
+          'date-example',
```
lennym: I think this would be a good candidate to be an example of the `parse` option on the Summary Page behaviour.

joefitter: yes, yes it would 👍

example/apps/example-app/translations/src/en/fields.json
lennym: I'd also possibly suggest that if labels are loaded from translations then they're from properties within the field itself. The idea of the component is that it treats the date as a single field, so adding more top-level properties to the fields translation file seems to be the opposite of what we're trying to do.

example/package.json
```diff
@@ -21,9 +21,11 @@
   },
   "author": "",
   "dependencies": {
+    "hof-behaviour-summary-page": "^2.0.0",
     "hof-bootstrap": "file:..",
-    "hof-controllers": "^4.1.1",
-    "hof-frontend-toolkit": "^1.0.0"
+    "hof-component-date": "^1.0.0",
+    "hof-frontend-toolkit": "^1.0.0",
+    "mixwith": "^0.1.1"
```
lennym: Is this actually a dependency?

lennym: Spoiler alert: no.

example/apps/example-app/index.js
joefitter: absolutely - I was planning on updating it once it's merged

example/package.json
joefitter: it was, but it aint no more

example/apps/example-app/index.js
lennym: Be the change you want to see in the world.

example/apps/example-app/translations/src/en/fields.json
joefitter: agree - this was one of those "I'll fix this later" bits

lennym: Yeah, I was just mentioning it as something to add to any issue you create.

example/package.json
lennym: Can we remove it then please?

joefitter: yep - I haven't pushed my changes yet

PR #127, opened 2017-02-17
Add setting for global and app-level behaviours

lennym:
This will allow behaviours to be mounted onto all controllers in an app or in a collection of apps.

Restores `baseController` setting, but with a deprecation warning.


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "hof-bootstrap",
   "description": "A bootstrap for HOF projects",
-  "version": "9.3.1",
+  "version": "10.0.0",
```
joefitter: How comes this was 9.3.1 - does it need a rebase?

PR #129, opened 2017-02-22
Allow apps to not have their own views directory

lennym:
If an app is only using common views, or views have been mounted elsewhere, then there is no need for an app to have it's own views directory. Allow apps which are created with no specified views directory to not throw if the views directory is the default and doesn't exist.


joefitter:
🎉

PR #130, opened 2017-03-02
Fix translations

lennym:
The previous implementation of translations didn't have the ability to support multiple languages because it did not specify the language anywhere in the translation method calls.

By moving the translations to middleware we can make use of built-in `Accept` header parsing to determine the user's desired language and translate accordingly (although we don't currently have content for any language other than `en`).

Additionally by automatically mounting translations from `hof-template-partials` we can simplify translation compilation in implementing projects as we remove any need to compile that content into translation files.

In particular this removes any need to define common translations in order to be able to load content for error pages and cookie/terms pages.

This is dependent on new versions of `hof-template-partials` and `hof-middleware` with the following merged:

* [x] https://github.com/UKHomeOfficeForms/hof-middleware/pull/24
* [x] https://github.com/UKHomeOfficeForms/hof-template-partials/pull/17
* [x] https://github.com/UKHomeOfficeForms/hof-template-mixins/pull/10

Posting now for early review.


lennym:
This isn't WIP anymore. I've tested with `npm link` into firearms and I can get it to display different "langauges" based on browser settings.

PR #131, opened 2017-03-14
Merge config instead of assigning

lennym:
This allows apps to set only parts of config objects (session, redis) without having to set the entire object.

### DIFF
--- a/index.js
+++ b/index.js
@@ -21,7 +21,7 @@ const customConfig = {};
 
 const getConfig = function getConfig() {
   const args = [].slice.call(arguments);
-  return Object.assign.apply(null, [{}, defaults, customConfig].concat(args));
+  return _.merge.apply(_, [{}, defaults, customConfig].concat(args));
 };
 
 const loadRoutes = (app, config) => {


index.js
```diff
@@ -21,7 +21,7 @@ const customConfig = {};
 
 const getConfig = function getConfig() {
   const args = [].slice.call(arguments);
-  return Object.assign.apply(null, [{}, defaults, customConfig].concat(args));
+  return _.merge.apply(_, [{}, defaults, customConfig].concat(args));
```
joefitter: do we need lodash here?

```js
return Object.assign.apply(Object, [{}, defaults, customConfig].concat(args));
```

joefitter: too late :(

lennym: Yes, that's literally the exact change I just made.

```
Object assign({ a: 1, b: 2 }, { c: 3}); // { c: 3 }
_.merge({ a: 1, b: 2 }, { c: 3}); // { a: 1, b:2, c: 3 }
```

In particular, it means we can set partial session properties from apps, without needing to specify the whole object, and therefore allows us to not have to re-implement the env var reading that bootstrap already does.

joefitter: ah yes, I didn't look at the removed code! :D wouldn't _.mergeDeep be more robust if combining things that could be multiple levels deep?

lennym: Maybe, if it existed.

`_.merge` is already a deep merge.

joefitter: shhhh

joefitter: it didn't used to be

joefitter: oh I'm thinking of cloneDeep. wotevs

lennym: You are probably thinking of `clone` which is shallow, and has a corresponding `cloneDeep`.

joefitter: 🍆

PR #132, opened 2017-03-14
Get the app to use hof-middleware-markdown

sulthan-ahmed:


index.js
```diff
@@ -138,6 +138,7 @@ function bootstrap(options) {
 
   app.use(userMiddleware);
   app.use(hofMiddleware.cookies());
+  app.use(require('hof-middleware-markdown')());
```
lennym: Move the require statement to the top of the file please.

sulthan-ahmed: Fair.

I looked at the firearms implementation that's why I did it like that.  
https://github.com/UKHomeOffice/firearms/blob/master/app.js#L33

I wasn't sure what was the best approach.  I guess require above keeps it consistent and maybe is the best practice

lennym: There are two differences:

* In firearms that code is at the top level of the file (i.e. not inside a function call), so will execute exactly once at start-up and no more.
* In firearms the containing file is short, and the require is on line 33, and so is easily visible - especially given that most of the firearms file is require statements already.

sulthan-ahmed: O yes I see that's in a function and it is short 👍 

I think when in doubt, I'll put it separately

lennym: Yeah, no-one ever went wrong putting all their `require`s at the top of the file.

The key is to always have them at the root of the module though. If you do that then you automatically resolve the full dependency tree in the first pass of the event loop, and you know that once your app has started then it won't later fail due to a missing module. And if it is going to fail then it will fail immediately on start-up.

By putting a require inside a function you open up the possibility that it might get called later, and error in the future. So keeping them all at the top is a nice pair of bullet-proof socks.

index.js
```diff
@@ -138,6 +139,7 @@ function bootstrap(options) {
 
   app.use(userMiddleware);
   app.use(hofMiddleware.cookies());
+  app.use(markdown());
```
lennym: We're almost 100% certainly going to want the markdown middleware to be loaded *before* the cookies and terms and conditions pages are routed in order to be able to use markdown for the content in those pages. I'd move the mounting above those routes.

lennym: Massive brain-fart. Somehow forgot I'd already merged this.

Don't worry, I'll do it.

sulthan-ahmed: yes good point about the markdown middleware being loaded before, I'll be interested to know where you move it

PR #138, opened 2017-05-12
Use the hof-logger

JoeChapman:


index.js
```diff
@@ -117,7 +117,7 @@ function bootstrap(options) {
     }
   });
 
-  if (config.env !== 'test' && config.env !== 'ci') {
+  if (config.env === 'production') {
```
lennym: I'd be tempted to get rid of this condition completely and use explicit log-level configuration for setting log levels.

lib/logger.js
```diff
@@ -1,61 +0,0 @@
...
-  const colors = {
-    info: 'green',
-    email: 'magenta',
-    warn: 'yellow',
-    error: 'red'
-  };
-
-  loggingTransports.push(
-    new winston.transports.Console({
-      level: config.loglevel,
```
lennym: This needs to be reflected in hof-logger.

lennym:
I agree 100% with the idea of this PR, but I really, really want to kill all the inference from `NODE_ENV` in favour of explicit config.

JoeChapman:
@lennym that's absolutely fair, good shout. I completely agree.

lennym:
While we're on the subject of logging. It'd be really good to have a middleware that adds a `req.log` method or something so that all the downstream behaviours and controllers can log things easily. 

I think we have trouble at the moment with a lot of functionality having no logging at all and so failures are a total black box, so being able to easily add `debug` level logging would be really helpful.

Probably out of scope for this PR though.

JoeChapman:
Yes, not in scope, but that is definitely something that is needed.

lennym:
For what it's worth, I think it's ok to infer a *default* log level from `NODE_ENV` if none is provided, but it needs to be independently configurable.

easternbloc:
@lennym 

> While we're on the subject of logging. It'd be really good to have a middleware that adds a req.log method or something so that all the downstream behaviours and controllers can log things easily.

> I think we have trouble at the moment with a lot of functionality having no logging at all and so failures are a total black box, so being able to easily add debug level logging would be really helpful.

> Probably out of scope for this PR though.

Churchill does this so if your project is using that it *should* be available...

lennym:
ORLY?



We should 100% use that thing then.

I suspect part of the reason we haven't is that we've been wrapping the request logger in a conditional so trying to rely on it downstream would throw errors in certain environments. That or people (a.k.a. me) just didn't know it was there.

lennym:
Meanwhile... needs more rebase.

lennym:
Is this one now redundant since you have your other one (#142)  @JoeChapman?

JoeChapman:
@lennym  I think so, I think so

easternbloc:
close the PR then?

PR #142, opened 2017-05-31
apply logger instance to request

JoeChapman:
So we can use the logger instance in any method invoked with a request object

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,8 +18,8 @@
   },
   "scripts": {
     "test": "npm run unit && npm run lint",
-    "unit": "NODE_ENV=test istanbul cover _mocha",
-    "unit:nocov": "NODE_ENV=test mocha",
+    "unit": "LOG_LEVEL=error istanbul cover _mocha",
+    "unit:nocov": "LOG_LEVEL=error mocha",
     "lint": "eslint .",
     "snyk": "snyk test",
     "snyk:dev": "snyk test --dev",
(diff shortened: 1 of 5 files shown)


lennym:
Churchill will automatically add a logger to the request as it is. So the code to overwrite `req.logger` is actually redundant, it's already there.

easternbloc:
word

JoeChapman:
@lennym @easternbloc can this be merged?

lennym:
Setting `LOG_LEVEL=silent` for tests seems like a better approach than hacking everywhere with proxyquire.

JoeChapman:
Using `proxyquire` because `logger` is invoked on each request and throwing 
```
Error: Transport already attached: console
      at Logger.add (node_modules/winston/lib/winston/logger.js:454:11)
      at node_modules/winston/lib/winston/logger.js:62:12
      at Array.forEach (native)
      at new exports.Logger (node_modules/winston/lib/winston/logger.js:61:24)
      at module.exports.config (lib/logger.js:9:1540)
      at bootstrap (index.js:9:6128)
      at Context.it (test/integration/behaviours.js:102:7)
      at Context.fn (node_modules/sinomocha/index.js:151:27)
      at Test.mocha.Test.run (node_modules/sinomocha/index.js:156:27)
```

package.json
```diff
@@ -18,7 +18,7 @@
   },
   "scripts": {
     "test": "npm run unit && npm run lint",
-    "unit": "NODE_ENV=test istanbul cover _mocha",
+    "unit": "LOG_LEVEL=error NODE_ENV=test istanbul cover _mocha",
     "unit:nocov": "NODE_ENV=test mocha",
```
lennym: We probably want the logs silenced for the `nocov` script too. And I don't think `NODE_ENV` needs to be set now.

lennym:
Can you bump `churchill` to latest please?

easternbloc:
@JoeChapman `1.1.1`

PR #143, opened 2017-06-01
Remove non-standard log levels

lennym:
I'm not aware of any need for the custom levels, but they make trying to do logging in a predictable way difficult because the expected log levels don't exist and so logging `debug` or `verbose` level messages fails silently.

Remove the custom log levels and use the default levels instead.

If there's a reason we do need those custom levels then we can add them back in but with the standard levels included.

### DIFF
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -3,24 +3,14 @@
 const winston = require('winston');
 
 module.exports = (config) => {
+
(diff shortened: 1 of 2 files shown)


easternbloc:
old code that's just festered around happy this is going tbh

PR #146, opened 2017-06-16
Change static server flag to be explicit config

lennym:
By inferring from the `NODE_ENV` as to whether to serve static assets in-app we are being over-opinionated about the infrastructure being used to run a hof service. In particular we are saying it is entirely necessary to have some kind of nginx proxy in front of *all* hof apps, which seems unnecessarily strict.

Instead allow the static middleware to be mounted in all cases unless explicitly disabled. In cases where a proxy is in place then the route will be effectively a no-op, since `/public` routes will be caught by nginx and should not hit the node app. However, the static mounting can still be disabled if desired by setting the appropriate config flag.


andymarke:
LGTM

PR #147, opened 2017-06-22
Update to hof-template-mixins to 4.2.2

sulthan-ahmed:


easternbloc:
package.lock file?

sulthan-ahmed:
I'm happy to add a package.lock file but @lennym has pointed out it has caused a few bugs meaning you have to delete it and do npm i again.  If a user wasn't aware of those bugs it could cause them some issues that they wouldn't necessarily know how to fix.

Is it essential and are we saying we should always have it and update it?

lennym:
I `.gitignore`-ed the package-lock file because we weren't using it for deployments (we don't use npm@5) and so maintaining it was a lot of effort for no benefit.

If we decide to update our deployments to npm@5 and maintain the package-lock file then it's fine to un-ignore it, but it's a separate decision to this PR.

Even if we do maintain the package-lock, the package.json should still reflect the actual dependencies - i.e. `hof-template-mixins@4.2.2` is *required* here to fix a bug (I presume - @sulthan-ahmed commit messages should explain the reasons for changes) and so should be specified in the package.json.

PR #149, opened 2017-07-05
Added ability to pass configuration from bootstrap configuration to markdown

LCairns:
Added code to allow the markdown configuration to be brought in through the bootstrap configuration file. This allows for the ability to set a default fallback language for markdown, the issue was brought to attention in acceptance tests we ran on our current HOF project where locale wasn't set in the browser and the application was producing an error.


lennym:
Published @ 12.1.2

Thanks @LCairns!

PR #158, opened 2017-11-22
Make static `pages` a 1st class citizen.

JoeChapman:
This enables users to define static `pages` at the same level
as `steps` either in an existing app adjacent to existing `steps`
or in a separate app without any `steps`.

Also deprecates `getCookies` and `getTerms`.

```js
        routes: [{
          views: `path/to/views`,
          steps: {
            '/one': {}
          },
          pages: {
            '/page-a': 'view-a',
            '/page-b': 'view-b'
          },
        }]
```


lib/router.js
```diff
@@ -15,11 +15,22 @@ function applyBehaviours(steps, behaviours) {
...
+  };
+  if (config.route.params) {
+    wizardConfig.params = config.route.params;
+  }
+  if (config.appConfig) {
+    wizardConfig.appConfig = config.appConfig;
+  }
+  return wizardConfig;
+}
+
```
lennym: You've fully broken #157!

lennym:
I was all set to ask for some README docs, but that doesn't seem to be a thing. Could you add something to hof-guide at your earliest convenience please?

JoeChapman:
> I was all set to ask for some README docs, but that doesn't seem to be a thing. Could you add something to hof-guide at your earliest convenience please?

yes I Iooked

PR #161, opened 2017-12-01
Prevent `disabled` csp directive being added

lennym:
When `csp.disabled === false` a directive of `disabled` was being added to the directives in the csp header. This caused a warning to be thrown in the browser. Prevent this from occurring.

### DIFF
--- a/test/integration/index.js
+++ b/test/integration/index.js
@@ -566,6 +566,25 @@ describe('bootstrap()', () => {
         });
     });
 
+    it('does not try to add a `disabled` directive', () => {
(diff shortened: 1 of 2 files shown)


test/integration/index.js
```diff
@@ -566,6 +566,25 @@ describe('bootstrap()', () => {
         });
     });
 
+    it('does not try to added a `disabled` directive', () => {
```
JoeChapman: small typo s/added/add

lennym: No there isn't... \*handwaves\*

PR #168, opened 2019-11-19
Add a readiness check

zuzak:
hof already has a "ping" check, that just ensures the app is alive.
This means that we can set up "liveness" checks with Kubernetes or
another platform.

This commit adds a "readiness" check too, which is a bit deeper.
It checks that the session store is accessible as well.


easternbloc:
1mo

easternbloc:
@zuzak is your npm username zuzak?

zuzak:
yup

easternbloc:
you've been added but npm caching is trolling me so your success of publishing may vary...

zuzak:
13.4.0 published ✅

PR #180, opened 2020-10-30
SD-50: Update winston logger to v3.3.3 and add improvements 

Alex-Swann:


lib/logger.js
```diff
@@ -26,23 +44,65 @@ module.exports = (config) => {
...
+    info: writeLogType('info'),
+    http: writeLogType('http'),
+    verbose: writeLogType('verbose'),
+    debug: writeLogType('debug'),
+    silly: writeLogType('silly')
+  };
 
-  return logger;
+  return Object.assign({}, allLogLevels, {
+    log: (level, ...args) => allLogLevels[level](...args)
```
Alex-Swann, line 106: @AlexFranklinHODDaT so Marvel created their own dependency called Churchill which hasn't been updated in 4 years. At the moment it sits in the middle of 'hof' as a dependency that orchestrates the logging. We need to look at it closer. But I've updated winston to the latest version and the use of `logger.log` is for version 2.x.x not 3.x.x. To get round this I've added this to hook into what it did before, (i.e. 1st arg is logLevel and the rest of args are the message). Need to create a card to either remove or migrate code from winston. Also maybe to ask Ben why they set it up in the 1st place. It might have been to make debugging HOF easier.

lib/logger.js
```diff
@@ -1,23 +1,41 @@
...
   const loggingTransports = [];
   const exceptionTransports = [];
   const notProd = config.env !== 'production';
 
+  const customFormatter = options => {
+    // - Return string will be passed to logger.
+    // - Optionally, use options.colorize(options.level, <string>) to
+    //   colorize output based on the log level.
+    return options.timestamp() + ' ' +
+      options.sessionID ? `sessionID=${options.sessionID}` : '' +
```
zuzak, line 16: I wonder if this would be neater?
```js
return [
    options.timestamp(),
    options.sessionID(),
    // ...
].filter(Boolean).join(' ')
```

sulthan-ahmed (commented):
As always your code is fab... 

Just a suggestion, especially as an opensource framework used by many, did you want to explain the improvements and how to use them for logging especially in your git commit?  If someone was to look at git log, I'm not sure RIL will mean much or explain the improvements tbh.  
I like this article about git commit that might help https://chris.beams.io/posts/git-commit/
This is good for PRs
https://www.pullrequest.com/blog/writing-a-great-pull-request-description/

lib/logger.js
Alex-Swann: Yep, this was copied over from RIL and imagine was setup most likely by a SO post by the looks of things. I wasn't sure to leave it in or not but went with the former to see if it got some thoughts. Will implement the above. Much nice.

lib/logger.js
Alex-Swann: Ah, it might have been more about getting it setup so the logger is accessible on the request object too when hof is brought into a project. Will make this conversation done and leave these comments here for future reference.

Alex-Swann:
@sulthan-ahmed yeah that article is good apart from the obvious fact it does not show you actually how to write a git commit message with a subject and body. Bit stupid tbh.

sulthan-ahmed (approved):

sulthan-ahmed:
good point @Alex-Swann I agree it would make life easier if he gave one

PR #183, opened 2020-12-07
SD-75: Address ZAP report issues to Improve Cookie Security

agraham19:
Addresses a few issues raised by an automated security scan:
* add sameSite: 'strict' attribute to session cookie
* disable caching
* set session cookie to httpOnly (prevents browser js reading it)

### DIFF
--- a/index.js
+++ b/index.js
@@ -117,6 +117,7 @@ const getContentSecurityPolicy = (config, res) => {
  *     @see {@link https://github.com/UKHomeOfficeForms/hof-middleware-markdown}
  * @param options.getTerms {boolean} Optional boolean - whether to mount the /terms endpoint
  * @param options.getCookies {boolean} Optional boolean - whether to mount the /cookies endpoint
+ * @param options.noCache {boolean} Optional boolean - whether to disable caching
  *
  * @returns {object} A new HOF application using the configuration supplied in options
  */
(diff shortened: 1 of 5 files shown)


index.js
```diff
@@ -143,6 +144,14 @@ function bootstrap(options) {
 
   app.use(helmet.noSniff());
 
+  if (config.noCache && config.noCache !== 'false') {
+    app.use((req, res, next) => {
+      res.setHeader('cache-control', 'no-store, no-cache, must-revalidate, proxy-revalidate');
```
zuzak, line 149: Would this be better as an array? https://nodejs.org/api/http.html#http_response_setheader_name_value

test/integration/index.js
```diff
@@ -734,6 +734,68 @@ describe('bootstrap()', () => {
...
+    it('should disable caching if noCache is true', () => {
+      const bs = bootstrap({
+        fields: 'fields',
+        routes: [{
+          views: `${root}/apps/app_1/views`,
+          steps: {
+            '/one': {}
+          }
+        }],
+        noCache: 'true'
```
robertdeniszczyc2, line 786: While this is still truthy, would it be more appropriate to have the boolean as type boolean rather than type string?

robertdeniszczyc2:
Confirming my comment has been addressed

zuzak (commented):
Looks good to me, but I don't have access to the Jira ticket to know for sure.

zuzak (approved):

PR #188, opened 2021-05-14
SD-139: Update hof session cookies to be secure when in Node producti…

Alex-Swann:
If in production mode, cookies should be secured.


MaxOSullivan (approved):
Looks good

PR #211, opened 2021-08-17
[MINOR]: Add html to pdf converter as a model api with documentation

Alex-Swann:


sulthan-ahmed (approved):
Great idea and super easy to implement now you've updated the docs too

PR #224, opened 2021-09-14
Update chalk dependency to fix high severity security issue

MaxOSullivan:
### What
Update 'chalk' dependency as it was a high severity security issue raised in the snyk scan


al01010:
I don't think you need to change the version number, as it will do it automatically when you merge it in.

MaxOSullivan:
> I don't think you need to change the version number, as it will do it automatically when you merge it in.

Is this documented anywhere?

al01010:
> > I don't think you need to change the version number, as it will do it automatically when you merge it in.
> 
> Is this documented anywhere?

In confluence, although but might be good to update the README.

PR #233, opened 2021-09-30
HOFF-158 added checkboxes

RobertMcCann:
What
Added checkbox page with validation if user does not select an option

How
Based off RIL form where user is asked for income

Why
Demonstrate checkboxes for the basic form

### DIFF
--- a/example/apps/example-app/index.js
+++ b/example/apps/example-app/index.js
@@ -30,6 +30,10 @@ module.exports = {
     '/address': {
       fields: ['building', 'street', 'townOrCity', 'postcode'],
       next: '/checkboxes',
+    },
+    '/checkboxes': {
+      fields: ['petsOwned'],
+      next: '/radiobuttons'
     }
   }
 };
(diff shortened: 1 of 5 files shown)


example/apps/example-app/translations/src/en/fields.json
```diff
@@ -27,5 +27,26 @@
    }, 
   "postcode": { 
     "label": "Postcode"
+  },
+  "incomeTypes" : {
+    "legend": "Select the options where you believe you receive income from",
```
al01010, line 32: Get rid of 'believe you'

example/apps/example-app/index.js
```diff
@@ -30,6 +30,11 @@ module.exports = {
     '/address': {
       fields: ['building', 'street', 'townOrCity', 'postcode'],
       next: '/checkboxes',
+    },
+    '/checkboxes': {
+      fields: ['incomeTypes'],
+      validate: ['required'],
```
al01010, line 36: I don't think you need this line here, this should only be in field.js

al01010:
Looks good to me

PR #245, opened 2021-10-06
HOFF-145 add phone number step

RobertMcCann:
What
Added phone number step to the form with validators
Why
Demonstrate phone number step in the form


example/apps/example-app/translations/src/en/fields.json
```diff
@@ -70,6 +70,10 @@
   "email" : {
     "label": "Email address"
   },
+  "phone": {
+    "label": "Phone number",
+    "hint": "International phone numbers require the international dialing code"
```
al01010, line 75: dialling

PR #263, opened 2021-11-16
HOFF-168 add feature testing for complex form

RobertMcCann:
What
Added feature testing and validation feature for complex form.

Why
Create feature testing for the complex form

### DIFF
--- a/example/test/_features/example_app/validation.feature
+++ b/example/test/_features/example_app/validation.feature
@@ -62,7 +62,7 @@ Feature: validations
     Then I should see 'Application sent' on the page
     Then I click the 'Start again' button
     Then I should be on the 'landing-page' page showing 'Choose one of the options below and press continue.'
-    
+
   @complex_form
   Scenario: Full Complex Form Submission
     Given I start the 'base' application journey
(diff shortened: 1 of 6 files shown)


example/test/_features/step_definitions/steps.js
```diff
@@ -82,3 +82,8 @@ Then('I should see the {string} error', async function (content) {
   await this.page.waitForSelector('body', { timeout: 15000 });
   expect(await this.page.innerText('.validation-summary.error-summary')).to.include(content);
 }.bind(World));
+
+Then('I select field {string} and value {string}', async function (field, value) {
```
al01010, line 86: On line 86 - I would remove the word, field and value before {string}, so that it is more consistent with the others, and reads a bit better above.

al01010: It also reads a bit better if you have it the other way round. so ... async function (value, field)

example/test/_features/example_app/validation.feature
```diff
@@ -62,4 +62,57 @@ Feature: validations
...
+    Then I should see the 'Select where the appeal hearing is to be held' error
+    Then I select 'England and Wales'
+    Then I click the 'Continue' button
+    Then I click the 'Continue' button
+    Then I should see the 'Enter details about why you are making a complaint' error
+    Then I fill 'complaintDetails' text area with 'I would like to make a complaint'
+    Then I click the 'Continue' button
+    Then I click the 'Continue' button
+    Then I should see the 'Select an appeal stage from the list' error
+    Then I select field 'appealStages' and value '01. First Tier IAC Appeal - In Country Appeals'
```
al01010, line 113: Alter line 113 to read as 'Then I select '01. First Tier IAC Appeal - In Country Appeals' from 'appealStages'. (note I have changed the word 'and' to 'from' as well)

PR #268, opened 2022-01-13
[MINOR]: Add default accessibility statement to the HOF framework whi…

Alex-Swann:
…ch is set to off by default


index.js
```diff
@@ -204,6 +204,16 @@ function bootstrap(options) {
   app.use(mixins());
   app.use(markdown(config.markdown));
 
+  if (config.getAccessibility === true) {
+    deprecate(
+      '`getAccessibility` option is deprecated and may be removed in future versions.',
+      'Use `pages` to define static cookies page.'
+    );
```
sulthan-ahmed, line 211: this is an interesting one because you are building this into HOF but on the other hand it could be deprecated.  So the ideal is to have a custom pages page?

frontend/template-partials/views/accessibility.html
```diff
@@ -0,0 +1,55 @@
+{{<layout}}
+
+  {{$propositionHeader}}{{/propositionHeader}}
+
+  {{$header}}
+    {{header}}
```
sulthan-ahmed, line 6: should this be `{{#t}}accessibility-header{{/t}}`? or something like that

Alex-Swann: No this works fine. But the title of the page is quite long. So going to make it 
```
{{#t}}base.accessibility{{/t}}
```
instead which is shorter otherwise the information in the tab is very long.

index.js
Alex-Swann: I've added the deprecate method because we do that for terms and cookies. It is just to make it consistent. It may be we keep these and just remove those deprecate messages at some point.

sulthan-ahmed (approved):
Fantastic work

frontend/template-partials/views/accessibility.html
Alex-Swann: Actually this works fine.

PR #271, opened 2022-01-14
HOFF-166 complex form country search select step

RobertMcCann:
What
Created country search select step for the complex form. Added required validation, labels and headers so it is similar to GRO. Added country select on summary page.
Added acceptance tests and unit test.
Added two dependencies via yarn:
jquery
typeahead-aria

Why
User can explore the search select component that is found in GRO.


example/apps/example-app/sections/summary-data-sections.js
```diff
@@ -33,5 +33,8 @@ module.exports = {
   ],
   complaintDetails: [
     'complaintDetails'
+  ],
+  countrySelect: [
+    'countrySelect'
   ]
```
al01010, line 39: I think on the summary page this would be better after address (as it is an address type question) or before complaints detail (as it comes before this question in the form)

example/apps/example-app/index.js
```diff
@@ -1,6 +1,7 @@
 /* eslint-disable */
 'use strict';
 
+const countrySelect = require('./behaviours/country-select')
```
al01010, line 4: Should this be CountrySelect to be consistent with the other behaviours?

example/apps/example-app/translations/src/en/validation.json
```diff
@@ -26,6 +26,10 @@
   "countryOfHearing": {
     "default": "Select where the appeal hearing is to be held"
   },
+  "countrySelect": {
+    "default": "Enter a valid country of residence",
+    "required": "Enter your country of residence"
+  },
```
al01010, line 32: Had a look at GRO and they use the following
`"countrySelect": {
    "default": "Enter your country",
    "equal": "Pick a country from the list"
  }`

example/apps/example-app/translations/src/en/pages.json
```diff
@@ -52,6 +55,9 @@
       "appealDetails": {
         "header": "Appeal details"
       },
+      "countrySelect": {
+        "header": "Country of Residence"
```
al01010, line 59: Residence should be in lower case

al01010:
Looks good to me

PR #276, opened 2022-01-24
HOFF-213 continue saved form step

al01010:
**What?**

Add continue-saved-form step to the complex form demo, as part of incorporating a demo to the example app. The page asks the user if they want to continue a form they saved previously. 

**Why**

Enable the user to eventually go back to a saved form saved under that email address.


sulthan-ahmed:
Hi @RobertMcCann do you know about this PR? Shall we close it?

PR #291, opened 2022-03-07
update urijs to resolve snyk vulnerability

Rhodine:
## What?
- bump urijs to 1.19.10
## Why?
- to resolve snyk vulnerability


al01010:
I think it is failing the tests as you have changed the yarn.lock file directly and it doesn't match what is in your package.json file. I think you need to change the version of urijs on the package.json file, and the run yarn, this will then automatically update the yarn.lock file

al01010:
Looks good to me

PR #295, opened 2022-03-16
Rename example to sandbox

sulthan-ahmed:
This is to distinguish between the example and sandbox apps and their purposes


.gitignore
```diff
@@ -6,43 +6,23 @@ coverage
 npm-debug.log
 .idea
 *.iml
+components/.DS_Store
```
Alex-Swann, line 9: you can just make this `.DS_Store` and then it will ignore it everywhere.

Alex-Swann (approved):
Just one comment

.gitignore
sulthan-ahmed: ok thanks updated

PR #297, opened 2022-03-22
HOF-25 - Added sanitsation middleware and appropriate tests

MCrawleyHomeOffice:
What?
In order to move away from using naxsi rules we need to sanitise the user input before saving it to session

Why?
Naxsi is a sledgehammer which we've had to disable many rules so our users don't get problems. Instead we're going to add new sanitisation rules that will manipulate the user input if they enter something we don't like to avoid possible SQL injection and other attacks.

How?
Added a new middleware in the base-controller called _sanitization() which will loop our user inputs and modify them accrdingly.

Testing?
Unit tests and linting tests.

Screenshots (optional)
N/A

Anything Else?
Rules for what we have agreed to do for each unwanted string are detailed on the original ticket - HOF-25.

### DIFF
--- a/config/hof-defaults.js
+++ b/config/hof-defaults.js
@@ -38,16 +38,17 @@ const defaults = {
   },
   serveStatic: process.env.SERVE_STATIC_FILES !== 'false',
   sanitisationBlacklistArray: {
-    //Input will be sanitised using the below rules
-    //The key is what we're sanitising out
-    //The regex is the rule we used to find them (note some dictate repeating characters)
-    //And the replace is what we're replacing that pattern with. Usually nothing sometimes a single character or sometimes a single character followed by a "-"
+    // Input will be sanitised using the below rules
+    // The key is what we're sanitising out
+    // The regex is the rule we used to find them (note some dictate repeating characters)
+    // And the replace is what we're replacing that pattern with. Usually nothing sometimes a
+    // single character or sometimes a single character followed by a "-"
     '/*': { regex: '\/\\*', replace: '' },
     '*/': { regex: '\\*\\/', replace: '' },
     '|': { regex: '\\|', replace: '' },
     '&&': { regex: '&&+', replace: '&' },
     '@@': { regex: '@@+', replace: '@' },
-    '/..;/': { regex: '/\\.\\.;/', replace: '' }, //Purposely input before ".." as they conflict
+    '/..;/': { regex: '/\\.\\.;/', replace: '' }, // Purposely input before ".." as they conflict
     '..': { regex: '\\.\\.+', replace: '.' },
     '/etc/passwd': { regex: '\/etc\/passwd', replace: '' },
     'c:\\': { regex: 'c:\\\\', replace: '' },
--- a/config/hof-defaults.js
+++ b/config/hof-defaults.js
@@ -36,7 +36,30 @@ const defaults = {
   apis: {
     pdfConverter: process.env.PDF_CONVERTER_URL
   },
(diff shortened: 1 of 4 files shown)


config/hof-defaults.js
```diff
@@ -36,7 +36,31 @@ const defaults = {
...
+    'c:\\': { regex: 'c:\\\\', replace: '' },
+    'cmd.exe': { regex: 'cmd\\.exe', replace: '' },
+    '<': { regex: '<', replace: '<-' },
+    '>': { regex: '>', replace: '>-' },
+    '[': { regex: '\\[+', replace: '[-' },
+    ']': { regex: '\\]+', replace: ']-' },
+    '~': { regex: '~', replace: '~-' },
+    '&#': { regex: '&#', replace: '&#-' },
+    '%U': { regex: '%U', replace: '%U-' }
+  }
```
sulthan-ahmed, line 63: naxsi stuff is private, should we do the same with this?

test/controller/spec/base-controller.spec.js
```diff
@@ -517,6 +517,43 @@ describe('Form Controller', () => {
...
+            values: {
+              value: value
+            }
+          };
+          form._sanitize(req, res, cb);
+          req.form.values.value.should.equal(expected);
+        });
+      });
+    });
+
```
sulthan-ahmed, line 556: should there be a test where you enter dodgy data, then it complains about it?

sulthan-ahmed (commented):
Were you able to test this on a service?

config/hof-defaults.js
Alex-Swann: Pfft, I'd say possibly not. Mainly because it would be a pain to test locally etc. I think this should be fine as naxsi bans traffic where in this case we are just escaping certain things. It might be worth putting this into its own config file though.

test/controller/spec/base-controller.spec.js
```diff
@@ -517,6 +517,43 @@ describe('Form Controller', () => {
...
+        { value: 'HELLO&#WORLD15', expected: 'HELLO&#-WORLD15' },
+        { value: 'HELLO%UWORLD16', expected: 'HELLO%U-WORLD16' },
+        {
+          value: '1/*2*/3|4&&5@@6..7/etc/PASSwd8C:\\9Cmd.eXe10/..;/11<12>13[14]15~16&#17%U18',
+          expected: '1234&5@6.7891011<-12>-13[-14]-15~-16&#-17%U-18'
+        }
+      ];
+
+      tests.forEach(({value, expected}) => {
+        it('sanitisation returns correct data', function () {
```
Alex-Swann, line 545: cool. Do we think it wise to have some tests for standard input to ensure we aren't accidentally changing inputs that normally wouldn't be changed?

config/hof-defaults.js
sulthan-ahmed: Good point 👍

MCrawleyHomeOffice: Thanks chaps, will move into its own config as suggested.

MCrawleyHomeOffice:
> Were you able to test this on a service?

Yes, it was tested on firearms and NRM using the acceptance tests.

test/controller/spec/base-controller.spec.js
MCrawleyHomeOffice: Happy paths will be tested as part of the existing acceptance tests so it might be duplication, but I'm happy to add a couple for completeness.

sulthan-ahmed (approved):
Great work @MCrawleyHomeOffice 👍

PR #301, opened 2022-03-28
Added field type check to sanitise middleware

MCrawleyHomeOffice:
Added sanitisation middleware to strip out unwanted/malicious data from user input.


controller/base-controller.js
```diff
@@ -162,6 +164,25 @@ module.exports = class BaseController extends EventEmitter {
     return validator(key, req.form.values[key], req.form.values, emptyValue);
   }
 
+  _sanitize(req, res, callback) {
+    // If we don't have any data, no need to progress
+    if(req.form.values) {
+      Object.keys(req.form.values).forEach(function (property, propertyIndex) {
+        // If it's not a string, don't sanitise it
+        if(_.isString(req.form.values[property])) {
```
Alex-Swann, line 172: might be good to have some unit tests to check for values passed to _sanitize in the req.form.values that aren't strings so if this was removed those tests would break

controller/base-controller.js
```diff
@@ -162,6 +164,25 @@ module.exports = class BaseController extends EventEmitter {
     return validator(key, req.form.values[key], req.form.values, emptyValue);
   }
 
+  _sanitize(req, res, callback) {
+    // If we don't have any data, no need to progress
+    if(req.form.values) {
```
Alex-Swann, line 169: Also a test to check for empty req.form.values in case this is removed too. This is always an array right?

MCrawleyHomeOffice: It's an object not an array so I will check for an empty object (ie. {}) as well as undefined.

controller/base-controller.js
MCrawleyHomeOffice: Good shout, done.

controller/base-controller.js
Alex-Swann: Maybe:
```
if(!_.isEmpty(req.form.values)) {
```

Alex-Swann: Then add a test for `req.form.values = {}` and make sure the unit tests do as they should.

Rhodine (approved):
lg

PR #308, opened 2022-04-14
Setup session rate limits

Alex-Swann:
Add rate limits that check a users session against redis to rate limit how many http requests they make to the form. This is to stop malicious spamming of a service, more specifically submissions being sent.


MCrawleyHomeOffice (approved):
Looks really good Alex. Nice work.

PR #312, opened 2022-04-26
Revert "Revert "Added field type check to sanitise middleware""

Alex-Swann:
Reverts UKHomeOfficeForms/hof#311


MCrawleyHomeOffice:
This PR is no longer needed as a new branch has been created for this work so we don't have an ugly "revert" branch name. Will keep for now and remove once validated that the new branch is fine.

PR #323, opened 2022-06-06
 UKVIC-168: DOB fields not highlighting in firefox when in edit mode.

SarahJaneLuff:
**Changes**	

- Changed form-focus.js to choose the first field when DOB is the only field on the page.
- Fixed dead link in validation message for date of birth group fields in validation.js


shamiluwais (approved):
looks good

PR #327, opened 2022-06-27
Update emailer component to handle signature 4 authentication

shamiluwais:
What
Pass updated transport options to Nodemailer create transport required for aws sig4 authentication.

Why
SMTP requires host(with AWS Region), port and authentication that have to be added to transport options in service config

How
Pass transport options instead of transport if host exists in the transportOptions.

Test
Manual testing in local and branch

### DIFF
--- a/components/emailer/emailer.js
+++ b/components/emailer/emailer.js
@@ -20,7 +20,12 @@ module.exports = class Emailer {
 
     const transport = transports[options.transport](options.transportOptions || {});
 
-    this.emailer = nodemailer.createTransport(transport);
+    if (options.transportOptions !== undefined &&
+      options.transportOptions.host !== undefined && options.transportOptions.host !== null) {
+      this.emailer = nodemailer.createTransport(options.transportOptions);
+    } else {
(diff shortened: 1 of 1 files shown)


sulthan-ahmed (commented):
Will there be a warning or error if you don't pass in these values?

components/emailer/emailer.js
```diff
@@ -20,7 +20,12 @@ module.exports = class Emailer {
 
     const transport = transports[options.transport](options.transportOptions || {});
 
-    this.emailer = nodemailer.createTransport(transport);
+    if (options.transportOptions !== undefined &&
```
Alex-Swann, line 23: The following would be much better here. Use lodash to get the attribute you need, if the attribute is undefined or null then it uses the transport variable instead. Much cleaner and refactors out all the unnecessary conditional code.
```
const emailerTransportOptions = _.get(options, 'transportOptions.host') || transport;
this.emailer = nodemailer.createTransport(emailerTransportOptions);
```

Alex-Swann (changes requested):
See comment

PR #334, opened 2022-08-04
add isWarning config for warning text in checkbox and radio fieldsets

Rhodine-orleans-lindsay:
## What 
- Added isWarning config to place warning text between the heading and checkbox/radio button form.
- Updated README
- Updated test for checkbox in mixins.spec.js to account for updated classNames
## Why
- Following the govuk design update page headings are included in the fieldset for single question checkboxes and radio pages. For templates that use {{render field}} this was causing warning text to appear above the page heading. 
- This change allows for placing warning text between the heading and checkbox/radio button form on single page questions where required
## How
- Added isWarning config to template-mixins.js
- Added govuk design system warning text component to checkbox-group.html and option-group.html
## Testing
- Tested locally


cmedlandho (approved):
Nice work Rhodine

PR #337, opened 2022-08-11
UKVIC-170: New Logic for whitelisting visited steps

SarahJaneLuff:
**Changes**

- check-progress.js changed to add new logic for calculating steps already visited in the service. 

**Why?**

- This is to prevent already visited steps from being invalidated and information lost on the confirmation page.

**Other Information**

- All current unit tests for forks in HOF pass.
- New logic has been manually tested on all services
- New acceptance test written for GRO and NRM to check journeys. 
- New acceptance tests written for UKVIC to replicate the initial bug of data being lost by incorrectly invalidated pages, all tests now pass.

### DIFF
--- a/wizard/middleware/check-progress.js
+++ b/wizard/middleware/check-progress.js
@@ -28,6 +28,41 @@ module.exports = (route, controller, steps, begin) => {
     return _.uniq(allSteps);
   };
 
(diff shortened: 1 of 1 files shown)


SarahJaneLuff:
Failing Tag checks

PR #340, opened 2022-08-16
add gov-uk class to template partials

RobertMcCann:
What
- Added gov-uk class to template partials
- commented out colour property on validation scss file

Reason
- class name invalid input was overriding gov uk design system with an older version
- colour property has been commented out in validation scss file because the error colour used is from another package and is overriding gov uk design system


Rhodine-orleans-lindsay (approved):
lg

PR #347, opened 2022-09-07
fix change cookie settings link

Rhodine-orleans-lindsay:
## What
- revert 'change your cookie settings' link to '/cookies'
## Why
- link was set as '/student-help/cookies' which was a non-existent page
## Testing
- tested locally


RobertMcCann:
looks good to me

PR #350, opened 2022-09-09
fix error link for input date

RobertMcCann:
What
Fixed issue with user unable to click validation summary which should link to input date field. This fix now follows more closely to the latest design system service update.

How
Checks to see if field.mixin is equal to input-date, if true it sets the field-day to the error link id which is where the user is taken when the user clicks on the validation summary. Replaced field.controlType with field.mixin as that property does not seem to exist as part of the form options fields.

Tested
Locally on example app, fields that are input date need to have mixin: input-date in fields.js

### DIFF
--- a/controller/controller.js
+++ b/controller/controller.js
@@ -177,7 +177,7 @@ module.exports = class Controller extends BaseController {
           // eslint-disable-next-line brace-style
           }
           // get first field for date input control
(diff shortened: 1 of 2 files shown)


Rhodine-orleans-lindsay (approved):
lgtm

sulthan-ahmed (approved):

PR #352, opened 2022-09-13
Error link for radio and checkbox + input-date mixin in sandbox

Rhodine-orleans-lindsay:
## What?
- add error link for checkbox as well as for radio and checkboxes where there is a toggle
- add input-date mixin to sandbox
## Why?
- Error link: There was no error link set for checkboxes. Also, where a radio or checkbox had a toggle, the error link did not highlight the first option as it should.
- Input-date mixin: to showcase the error link functionality for dates in the sandbox
## How?
-  Error link: added if block where the first option of a radio/checkbox option was an object due to having toggle.
- Input-date mixin: added mixin to fields.js
## Testing?
Tested locally


RobertMcCann:
looks good to me

PR #353, opened 2022-09-14
remove govuk-radios__hint class from checkbox group

RobertMcCann:
What?
* Removed class 'govuk-radios__hint' from checkbox group 

Why?
* Class 'govuk-radios__hint' was adding unnecessary padding/margin to option hints for checkboxes and is not in design system service. 
* Design system service shows that it should only contain class 'govuk-hint'


Rhodine-orleans-lindsay (approved):
lgtm

PR #355, opened 2022-09-21
GOVUK update fix summary style

Rhyad-Zergane-HO:
* Update summary-table-row.html
* As it used old styling for a link updated to govuk-link


sulthan-ahmed (approved):
nice work

PR #357, opened 2022-09-26
HOFF-45: Change heading to size l

Rhyad-Zergane-HO:
* Update maincontent-left.html
* Update package.json
* To ensure the versio can be released and to ensure heading size is correct


Rhodine-orleans-lindsay (approved):
lgtm

PR #363, opened 2022-10-06
remove tab border from cookie banner

Rhodine-orleans-lindsay:
## What 
- Remove tab border from cookie banner
## Why 
- The border around the cookie banner created when clicking tab was oddly shaped. Following the govuk design system guidelines, there is no tab index set for the cookie banner div so this border should not be there.
## How
- Removed the tab index in cookie-banner.html
## Testing
- Tested locally

### DIFF
--- a/frontend/template-partials/views/partials/cookie-banner.html
+++ b/frontend/template-partials/views/partials/cookie-banner.html
@@ -18,7 +18,7 @@
 				<a class="govuk-link" href="/cookies">View cookies</a>
 		</div>
 	</div>
-	<div class="gem-c-cookie-banner__confirmation govuk-width-container" tabindex="0" hidden=""  id="cookie-banner-submitted" >
+	<div class="gem-c-cookie-banner__confirmation govuk-width-container" hidden=""  id="cookie-banner-submitted" >
 		<p class="gem-c-cookie-banner__confirmation-message" role="alert">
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "hof",
   "description": "A bootstrap for HOF projects",
-  "version": "20.0.1",
```
sulthan-ahmed, line 4: I don't believe you need to up the version number, git actions will do it for you when it is merged

frontend/template-partials/views/partials/cookie-banner.html
```diff
@@ -18,7 +18,7 @@ <h2 class="govuk-cookie-banner__heading govuk-heading-m">Cookies on {{#appName}}
 				<a class="govuk-link" href="/cookies">View cookies</a>
 		</div>
 	</div>
-	<div class="gem-c-cookie-banner__confirmation govuk-width-container" tabindex="0" hidden=""  id="cookie-banner-submitted" >
+	<div class="gem-c-cookie-banner__confirmation govuk-width-container" hidden=""  id="cookie-banner-submitted" >
```
sulthan-ahmed, line 21: nice

sulthan-ahmed (commented):
lgtm, I would just remove the version otherwise good to merge

package.json
Rhodine-orleans-lindsay: thanks Sulthan - I've rebased following Alex's changes so this change should be removed now

PR #367, opened 2022-10-20
HOFF-45 fix layout footer order

sulthan-ahmed:
## What

fix layout footer order

## Why

* Terms should be before accessibility according to rotm, firearms and others on uat
* The Govuk design broke that order.  This fixes it


shamiluwais (approved):
lgtm

sulthan-ahmed:
npm test for version 12 failed for a redis version. Test pass with other versions and version 14 is the one we work on so will merge this

PR #368, opened 2022-10-20
Fix github actions stop support for node 10 & 12

sulthan-ahmed:
## What

Fix github actions stop support for node 10 & 12

## Why

* For some unknown reasons, node v12 redis 4 fails unit tests
* node v10, v12 is not supported anymore so no need to run tests for these


Alex-Swann (approved):

sulthan-ahmed:
thanks @Rhyad-Zergane-HO it's just I didn't want to wait while node 12 and redis 4 keeps breaking tests.  It slows everything down 😄

PR #371, opened 2022-11-21
HOFF-316: Checkbox and Radio Button Fix for when Java Script is disabled

SarahJaneLuff:
**Changes**

- panel.html changed to allow for radio buttons and checkboxes to function correctly with dependencies when javascript is disabled in the browser.
- HOF version changed to 20.1.13


shamiluwais (approved):
looks good

PR #374, opened 2022-11-30
HOF-318: Dependency validation fix for checkbox-groups with multiple dependencies per option

SarahJaneLuff:
**Changes**

- validationLink created to make a child field link back to it parent option in checkbox-groups. This is to remedy incorrect validation errors being generated for mulitple child dependency fields when the parent option was not selected.
- HOF updated to 20.1.16


shamiluwais (approved):
looks good

PR #376, opened 2022-12-09
Extend SASS config options in hof-build

johndallen:
The current implementation of hof-build doesn't allow for CSS output to be compressed or warnings to be suppressed (they cause a lot of noise in the console output). This PR adds optional configuration to apply these 2 options via hof.settings, anyone not choosing to configure either of them will see the default behaviour ie. not compressed CSS output and deprecation warnings showing in the console output when compiling SASS.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "hof",
   "description": "A bootstrap for HOF projects",
-  "version": "20.1.17",
+  "version": "20.2.0",
   "license": "MIT",
   "main": "index.js",
   "author": "HomeOffice",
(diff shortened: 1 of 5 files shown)


sulthan-ahmed:
Thanks for the PR, we just need to test it on and off and on multiple services

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "hof",
   "description": "A bootstrap for HOF projects",
-  "version": "20.1.17",
```
sulthan-ahmed, line 4: Can you bump this to a minor version.  We use semantic versioning and I'd say this is adding functionality rather than a patch.

Even if it didn't we try to now do minor updates and have our services use tilde instead of carets so minor updates won't effect us going forward unless we update our versioning

sulthan-ahmed:
Also, can you add an explanation of how to use this feature in the README

PR #384, opened 2023-02-16
HOFF-324: create time mixin

Rhodine-orleans-lindsay:
## What
- Add time mixin that can be used across services

## Why
- This was required by a new service but would be more useful in the hof framework so it can be used by any other services that require it.

## How
- Create time component
- Create time input view
- Configure template mixin
- Add relevant validators
- Add relevant styling
- Add sandbox example
- Add relevant tests

## Testing
- Manual tessting
- Automated tests passing locally


components/time/index.js
```diff
@@ -0,0 +1,172 @@
...
+  _.mapKeys(_.pick(body, Object.keys(fields)), (value, fieldKey) =>
+    fieldKey.replace(`${key}-`, '')
+  );
+
+// accepts a time value in the format kk-mm and fields config,
+// returns a map of key: value pairs for the intermedate fields
+const getPartsFromTime = (time, fields) =>
+  time.split(':')
+    .slice()
+    // .reverse()
```
sulthan-ahmed, line 26: did you mean to have this comment?

Rhodine-orleans-lindsay: I've removed this comment now

sulthan-ahmed: ta

PR #392, opened 2023-07-14
Feature/acq 132 option hint aria

SarahJaneLuff:
**Changes**

- Bug fix for ASC, where hint texts are missing and Aria tags are erroring.


jamiecarterHO (approved):
lgtm

PR #419, opened 2024-02-04
HOFF-593- add more information to 404 error page not found

TemitopeAyokuHO:
## What 
[HOFF-593](https://collaboration.homeoffice.gov.uk/jira/browse/HOFF-593) Adding content to 404 error page
## Why
To follow the GDS standard for 404 error page 
## How 
-  research the design system from [Gov.uk](https://design-system.service.gov.uk/patterns/page-not-found-pages)
-  added content to 404.html page

## Test
Tested locally

### DIFF
--- a/frontend/template-partials/views/404.html
+++ b/frontend/template-partials/views/404.html
@@ -1,9 +1,25 @@
 {{<layout}}
-  {{$header}}
-    {{title}}
-  {{/header}}
+  
+  {{$header}}  {{title}}  {{/header}}
+  
   {{$content}}
-    <p>{{description}}</p>
-    <a href="/" class="button" role="button">{{#t}}buttons.start-again{{/t}}</a>
+    <div class="govuk-width-container">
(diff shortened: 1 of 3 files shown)


frontend/template-partials/views/404.html
```diff
@@ -9,13 +9,13 @@
           <div class="govuk-grid-column-two-thirds">
             <h1 class="govuk-heading-l">{{title}}</h1>
               <p class="govuk-body">
-              {{paragraph1}}
+                If you typed the web address, check it is correct.
```
sulthan-ahmed, line 12: I believe rather than adding the content directly, you want to add it into the translation for paragraph1, paragraph2, paragraph 3

TemitopeAyokuHO: TBH i did that previously and i thought the we follow the same approach with the 500 error page

sulthan-ahmed: Translations are better because then it can be in different languages.

PR #434, opened 2024-03-21
HOFF-661: Deindex ETA webform

mislam987:
## What
De-index the ETA webform - [HOFF-661](https://collaboration.homeoffice.gov.uk/jira/browse/HOFF-661) 

## Why
So that it is not returned in browser searches

## How
- Added robots meta tag to /frontend/govuk-template/govuk_template.html, with the content as "noindex"

## Test
Tested locally

### DIFF
--- a/frontend/govuk-template/build/govuk_template.html
+++ b/frontend/govuk-template/build/govuk_template.html
@@ -4,6 +4,7 @@
 <!--[if gt IE 8]><!--><html lang="{{ htmlLang }}" class="govuk-template"><!--<![endif]-->
   <head>
     <meta charset="utf-8" />
+    <meta name="robots" content="noindex">
     <title>{{ pageTitle }}</title>
     {{{ head }}}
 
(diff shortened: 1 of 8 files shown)


frontend/govuk-template/build/govuk_template.html
```diff
@@ -4,6 +4,7 @@
 <!--[if gt IE 8]><!--><html lang="{{ htmlLang }}" class="govuk-template"><!--<![endif]-->
   <head>
     <meta charset="utf-8" />
+    <meta name="robots" content="noindex">
```
sulthan-ahmed, line 7: So this is for ETA and HOF is for all of our services, NHS and multiple other teams.  By implementing it this way you will impact all the teams

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "hof",
   "description": "A bootstrap for HOF projects",
-  "version": "20.4.2",
+  "version": "20.4.3",
```
sulthan-ahmed, line 4: Are you looking to publish a test version i.e. beta?you wouldn't want to publish this for the proper version of HOF as it'll impact everyone.

PR #440, opened 2024-04-02
HOF 592 Added 403 Error page (Forbidden)

TemitopeAyokuHO:
## What

[HOFF-592](https://collaboration.homeoffice.gov.uk/jira/browse/HOFF-592)- HTTP Error 500 (Internal Server Error) Error Page

## Why

To display an Forbidden error page to users, which follows the Government Design Standard (GDS)

## How 
- Added html content in 403.html page to display 403 forbidden error message
- Added 403 text in errors.json file.
- Modify the previous status code assigned to 403

## Testing
- Wrote unit test to unsure error page pass.


serviceops:
YOUR MAIL HAS NOT BEEN DELIVERED

The email you have sent to ***@***.*** is for a legacy domain which has been deleted.  Your mail has not been delivered.

Please update your address book and re-send your mail to the same address but having removed .gsi. from the domain part.

e.g. instead of sending to ***@***.*** please re-send your mail to ***@***.***

PR #441, opened 2024-04-02
HOFF 591 Added 401 error page for Unauthorised

TemitopeAyokuHO:
## What

[HOFF-591](https://collaboration.homeoffice.gov.uk/jira/browse/HOFF-591)- HTTP Error 401 (UnAuthorised) Error Page

## Why

To display an Unauthorised error page to users, which follows the Government Design Standard (GDS)

## How 
- Added html content in 401.html page to display 401 Unauthorised error message
- Added 401 text in errors.json file.
- Modify the previous status code assigned to 401 which was used for session timeout replaced with 408 status code

## Testing
- Wrote unit test to unsure error page pass.

### DIFF
--- a/frontend/template-partials/translations/src/en/errors.json
+++ b/frontend/template-partials/translations/src/en/errors.json
@@ -11,6 +11,11 @@
     "title": "Page not found",
     "description": "This page does not exist"
   },
+  "401": {
+    "title": "Unauthorised",
+    "description": "Unauthorise user",
+    "paragraph": "You do not have access to this page."
+  },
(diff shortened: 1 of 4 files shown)


serviceops:
YOUR MAIL HAS NOT BEEN DELIVERED

The email you have sent to ***@***.*** is for a legacy domain which has been deleted.  Your mail has not been delivered.

Please update your address book and re-send your mail to the same address but having removed .gsi. from the domain part.

e.g. send to @homeoffice.gov.uk.

PR #465, opened 2024-08-12
HOFF-403 - Update pkgs with security vulnerabilities

shamiluwais:
**What?**

Update security vulnerabilities - [HOFF-403](https://collaboration.homeoffice.gov.uk/jira/browse/HOFF-403)

**Why?**

Update vulnerabilities that are flagged up by trivy in file-vault

**How?**
Updated express-partial-templates and notifications-node-client to later versions that did not have vulnerabilities
 
**Testing?**

Updated file-vault module with the hof beta with the above fixes and the built file-vault image tested using ROTM service

**Other**
fs-tree-traverse npm has a vulnerability but cannot be resolved as it in a private npm space: https://www.npmjs.com/package/fs-tree-traverse


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "hof",
   "description": "A bootstrap for HOF projects",
-  "version": "21.0.3",
+  "version": "21.0.12-upgrade-vuln-pkgs-beta",
```
william-gu-hof, line 4: why 12-upgrade-vuln-pkgs-beta? what's between .3 and .12? why is this beta? are there more work to be done?

package.json
```diff
@@ -53,7 +53,7 @@
     "duplexify": "^3.5.0",
     "express": "^4.17.1",
     "express-healthcheck": "^0.1.0",
-    "express-partial-templates": "^0.2.0",
+    "express-partial-templates": "^0.2.1",
```
william-gu-hof, line 56: Are these 2 packages updates done because you know it is saved to do after doing the upgrade in file-vault? As file vault is usually running as a separate instance

william-gu-hof (approved):
Just a couple of questions added to the package.json file

package.json
shamiluwais: We publish a beta to be reviewed and tested with hof services before publishing a new version, 12 was the next available for beta version

package.json
shamiluwais: There were vulnerabilities in these packages which were flagged up in the file-vault pipeline

PR #475, opened 2024-09-05
HOFF-672: Add de-indexing toggle functionality

mislam987:
## What? 
Add feature to enable the toggle of de-indexing so that it can be set per service ([HOFF-672](https://collaboration.homeoffice.gov.uk/jira/browse/HOFF-672))

## Why? 
- To allow each service to be de-indexed or not, depending on the requirements
- To not return the webpage in browser searches where forms are de-indexed

## How
- Add variable in config/hof-defaults.js to pick up the de-index value from each service via the env variable, or setting the default value as false
- Add lib/deindex.js to retrieve the value of the de-index variable from the config file and set a variable to be used in the head.html partial
- Add conditional in frontend/template-partials/views/partials/head.html to de-index the form if the value is true

## Testing?
Tested locally in services and in branch using a beta version

## Check list
- [x] Ensure drone builds are green especially tests


muthrajan (approved):
Looks ok to me

PR #488, opened 2024-12-12
Support +447 and 07 for ukmobilephone validator

robertdeniszczyc2:
Updates the ukmobilephone validator to support numbers starting with 07 or +447


sulthan-ahmed:
Looks good, I just need to publish, if I publish a beta package, could you pull in and test it on your service?

sulthan-ahmed:
previous beta package had a bug from session-timeout so published a new one from the latest version of hof, hopefully, this will work, 
`hof@22.2.3-ukmobilephone-validator-beta`

PR #492, opened 2025-01-09
Hoff 1059 fix functional test

Rhodine-orleans-lindsay:


Adesh-Ramchurn (commented):
Hi @Rhodine-orleans-lindsay just a small comment, there are some tests that are being skipped. Are they required?

Rhodine-orleans-lindsay:
> Hi @Rhodine-orleans-lindsay just a small comment, there are some tests that are being skipped. Are they required?

Hi @Adesh-Ramchurn, no they are not required here. Those are auto-tags which only run on the master branch - see https://github.com/UKHomeOfficeForms/hof/blob/977f7bf988a6586b7e179ef35bc40354c7961730/.github/workflows/automate-tag.yml#L48

Adesh-Ramchurn (approved):
Thanks for clarifying that Rhodine. Code looks good

PR #523, opened 2025-05-12
WIP: HOFF-1195 - maxlengthAttribute unit tests

vivekkumar-ho:
## Check list


test/frontend/mixins.spec.js
```diff
@@ -599,6 +599,66 @@ describe('Template Mixins', () => {
...
+        middleware(req, res, next);
+        res.locals['input-text']().call(res.locals, 'field-name');
+        render.should.have.been.calledWith(
+          sinon.match({
+            maxlengthAttribute: false
+          })
+        );
+      });
+
+      it('sets maxlengthAttribute to false when set to an invalid value in field config', () => {
```
sulthan-ahmed, line 646: Hey man, I noticed you used the word "invalid".  Would it be more precise to use the word "non-boolean" as 'true' is actually a valid string in JS.   Technically, I don't think it's "invalid" from a syntactical or data-type perspective. Instead, I think what's happening here is the functionality being tested expects a boolean value, and 'true' is a non-boolean value.  What do you think?

test/frontend/mixins.spec.js
```diff
@@ -599,6 +599,66 @@ describe('Template Mixins', () => {
...
+        res.locals.options.fields = {
+          'field-name': {
+            maxlengthAttribute: 'true'
```
sulthan-ahmed, line 649: I noticed the test uses 'true' as an example of an invalid value for maxlengthAttribute. While this works, it might not be immediately clear that it's being treated as a non-boolean value. Maybe using something like a number like 123 could make the intention behind the test even clearer and easier for others to understand. What do you think?

test/frontend/mixins.spec.js
vivekkumar-ho: yes that makes sense. I was thinking to add something like (e.g. string) but I removed it. Will add that to make it clear.

test/frontend/mixins.spec.js
vivekkumar-ho: Thanks @sulthan-ahmed I've updated the test to run on multiple non boolean values.

test/frontend/mixins.spec.js
sulthan-ahmed: nice one man

sulthan-ahmed:
I really like your What and why, do you think it be worth adding it as part of your commit for version history?

test/frontend/mixins.spec.js
```diff
@@ -599,6 +599,71 @@ describe('Template Mixins', () => {
       });
     });
 
+    describe('maxlengthAttribute - input-text', () => {
```
sulthan-ahmed, line 602: I think it might be safer to reset `render`, `render.should.have.been.calledWith` Otherwise, previous calls may interfere.
This might not work because I think `render.reset()` is in the older sinon versions before each test case
```suggestion
    describe('maxlengthAttribute - input-text', () => {
      beforeEach(() => {
      // Reset the render spy/mock before each test
      render.resetHistory();
  });
```

test/frontend/mixins.spec.js
```diff
@@ -599,6 +599,71 @@ describe('Template Mixins', () => {
...
+          middleware(req, res, next);
+          res.locals['input-text']().call(res.locals, 'field-name');
+          render.should.have.been.calledWith(
+            sinon.match({
+              maxlengthAttribute: false
+            })
+          );
+        });
+      });
+    });
```
sulthan-ahmed, line 665: I'm worried the forEach test for non-boolean values runs all cases in a single test. This can make it harder to pinpoint which input failed if the test breaks.

I think this makes each non-boolean case is its own test, so failures are easy to trace.
```suggestion
    // List of non-boolean values to test
      const nonBooleanTypeValues = [
        'string', null, undefined, 1234, {}, []
      ];
      nonBooleanTypeValues.forEach(nonBooleanValue => {
        it(`should set maxlengthAttribute to false when set to non-boolean value (${JSON.stringify(nonBooleanValue)}) in field config`, () => {
          res.locals.options.fields = {
            'field-name': {
              maxlengthAttribute: nonBooleanValue
            }
          };
          middleware(req, res, next);
          res.locals['input-text']().call(res.locals, 'field-name');
          render.should.have.been.calledWith(
            sinon.match({ maxlengthAttribute: false })
          );
        });
      });
    });
```

sulthan-ahmed:
Great work @vivekkumar-ho 
These are just minor suggestions.  

Sorry it took so long to review

sulthan-ahmed:
also, what i wrote might not work but you get the gist of it hopefully 🙂

test/frontend/mixins.spec.js
vivekkumar-ho: Thanks Sulthan. since its a stub, I've used resetHistory now.

vivekkumar-ho:
> I really like your What and why, do you think it be worth adding it as part of your commit for version history?

While merging, shall we choose 'Squash and merge" and add a brief description?

sulthan-ahmed (approved):
Great work @vivekkumar-ho

PR #528, opened 2025-05-29
HOFF-1204: Allow Service unavailable text to use html tags

Rhodine-orleans-lindsay:
## What? 
Allow Service unavailable text to use html - [HOFF-1204](https://collaboration.homeoffice.gov.uk/jira/browse/HOFF-1204)
## Why? 
So that elements like links can be included in text
## How? 
Amended the service-unavailble html
## Testing?
Tested in sandbox. Tests passing locally
## Check list

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2025-05-29, Version 22.7.2 (Stable), @Rhodine-orleans-lindsay
+### Fixed
+- Service unavailable text can now use html
+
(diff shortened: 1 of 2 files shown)


vivekkumar-ho (approved):
Looks good to me. {{{ returns unescaped HTML, this can potentially introduce XSS type of vulnerabilities, however, since this is not driven by user input and coming from a pre defined content at our end, we should be fine.

PR #548, opened 2025-09-03
HOFF-1265: Back button leads to session timeout screen missing text

Rhodine-orleans-lindsay:
## What? 
HOFF-1265: Back button leads to session timeout screen missing text - [HOFF-1265](https://collaboration.homeoffice.gov.uk/jira/browse/HOFF-1265)
## Why? 
- When the session initially times out everything on the page is working as expected, clicking the browser's back button on the initial timeout page redirects to a blank timeout page with no content.The Start Again button on this page is non-functional.
## How? 
- amended index.js to get the session timeout page and show session-timeout content if there is a session timeout error and to show 'page not found' if there is no session timeout error
- removed '/session timeout' step from sandbox's apps/sandbox/index.js as this is now handled in hof/index.js
- updated changelog
## Testing?
- tests passing locally
- beta version tested on ARE, VIPTT and ASC using hof@22.9.0-fix-timeout-back-button-bug-beta.4
### Before fix
https://github.com/user-attachments/assets/3a6ec728-1ed0-4cb3-aec1-9e49cc503b7a

### After fix
https://github.com/user-attachments/assets/a7e01350-b89d-4865-9108-b2c673fc7b88

## Check list

### DIFF
--- a/index.js
+++ b/index.js
@@ -249,6 +249,30 @@ function bootstrap(options) {
       return next();
     });
   }
+
+  /**
+   * Handles requests to the session timeout page.
+   * - If the user has a session cookie but their session is missing or inactive,
+   *   this triggers a session timeout error handled by error middleware.
+   * - Otherwise, responds with a 404 "Page Not Found" error.
+   * This route ensures the timeout page only appears after an actual session expiry.
+  */
+  app.get('/session-timeout', (req, res, next) => {
+    if ((req.cookies['hof-wizard-sc']) && (!req.session || req.session.exists !== true)) {
+      const err = new Error('Session expired');
+      err.code = 'SESSION_TIMEOUT';
+      return next(err);
+    }
+    const err = new Error('Not Found');
+    err.status = 404;
+    const locals = Object.assign({}, req.translate('errors'));
+    if (locals && locals['404']) {
+      return res.status(404).render('404', locals['404']);
+    }
+    // Fallback: render a basic 404 page if translation is missing
+    return res.status(404).send('Page Not Found');
+  });
+
   if (config.getAccessibility === true) {
     deprecate(
       '`getAccessibility` option is deprecated and may be removed in future versions.',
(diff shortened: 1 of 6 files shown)


index.js
```diff
@@ -249,6 +249,19 @@ function bootstrap(options) {
...
+  app.get('/session-timeout', (req, res, next) => {
+    if (req.cookies['hof-wizard-sc'] && req.session.exists !== true) {
+      const err = new Error('Session expired');
+      err.code = 'SESSION_TIMEOUT';
+      return next(err);
+    }
+    const err = new Error('Not Found');
+    err.status = 404;
+    const locals = Object.assign({}, req.translate('errors'));
+    return res.status(404).render('404', locals['404']);
```
sulthan-ahmed, line 262: It might be worth checking if locals['404'] exists before using it.  By adding this code, I think it stops a blank page if translation data is missing.
```suggestion
  const locals = Object.assign({}, req.translate('errors'));
  if (locals && locals['404']) {
    return res.status(404).render('404', locals['404']);
  } else {
    // Fallback: render a basic 404 page if translation is missing
    return res.status(404).send('Page Not Found');
  }
```

sulthan-ahmed:
You could add some jest tests e.g.
```
it('shows the session timeout error when the user has a session cookie but no active session', () => {
     const req = {
      cookies: { 'hof-wizard-sc': true },
      session: null
    };
// add the function and expect bit
....
it('shows a "Page Not Found" error when the session is active or not missing', () => {
    const req = {
      cookies: {},
      session: { exists: true },
      translate: () => ({ '404': { message: 'Not Found' } })
    };
// add the function and expect bit
```

index.js
```diff
@@ -249,6 +249,19 @@ function bootstrap(options) {
       return next();
     });
   }
+
```
sulthan-ahmed, line 252:
```suggestion
/**
 * Handles requests to the session timeout page.
 * - If the user has a session cookie but their session is missing or inactive,
 *   this triggers a session timeout error handled by error middleware.
 * - Otherwise, responds with a 404 "Page Not Found" error.
 * This route ensures the timeout page only appears after an actual session expiry.
 */
```

sulthan-ahmed:
> > Great work, I think this is a separate question but I think it's worth asking. When a user clicks on start again, are we expecting them to go somewhere else or stay on the same page
> 
> Via the error middleware, the 'Start again' button goes to the first step of a form - so either '/' or '{{baseUrl}}/'

Just curious does that work, I couldn't see in the video.

PR #655, opened 2026-05-14
HOFF-2085: Bug fix - some service level views not overriding framework views

Rhodine-orleans-lindsay:
## What? 
[HOFF-2085](https://collaboration.homeoffice.gov.uk/jira/browse/HOFF-2085) - fix bug where some service level views not overriding framework views
## Why? 
When an service uses a custom view to override the default framework equivalent e.g. layout.html, the framework view is still being used. This happens for views that are not associated with a step e.g. layout.html, page.html. A custom confirmation.html view for the '/confirmation' step correctly overrides the framework default. 
## How? 
- refactored the lib/settings.js file so that view precedence is `service level <route>/views -> service level common/views -> hof template-partials/views` to allow services to override equivalent framework views correctly
-  set res.locals.routeConfig in lib/router.js so it can be used in lib/settings.js
- updated how template-mixins.js and mixins test call nunjucks environment after changes to lib/settings
## Testing?
tested locally in hof/sandbox and  locally in paf using beta version 24.0.0-nunjucks-refactoring-fix-view-override-bug.beta.12
## Anything Else? (optional)
- fixed incorrect syntax in readme
## Check list

### DIFF
--- a/lib/settings.js
+++ b/lib/settings.js
@@ -1,6 +1,5 @@
 'use strict';
 
-const _ = require('lodash');
 const fs = require('fs');
 const path = require('path');
 const nunjucks = require('nunjucks');
@@ -37,7 +36,10 @@ module.exports = async (app, config) => {
   app.enable('view cache');
 
   if (config.views) {
-    const viewsArray = _.castArray(config.views);
+    const viewsArray = Array.isArray(config.views)
+      ? config.views
+      : [config.views];
+
     viewsArray.slice().reverse().forEach(view => {
       const customViewPath = path.resolve(config.root, view);
       try {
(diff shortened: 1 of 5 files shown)


lib/settings.js
```diff
@@ -52,67 +52,152 @@ module.exports = async (app, config) => {
...
-        const fullPath = path.join(dir, entry.name);
-
-        if (entry.isDirectory()) {
-          if (entry.name === 'views') {
-            results.push(fullPath);
-          } else {
-            addDir(fullPath);
-          }
-        }
+  const isLocalDev = process.env.NODE_ENV === 'local';
```
gregwolversonHO, line 57: What's the difference here between `local` and the usual `development` modes?

lib/settings.js
```diff
@@ -52,67 +52,152 @@ module.exports = async (app, config) => {
...
+      : path.resolve(config.root || process.cwd(), viewPath);
+  };
+
+  const getRouteViewPaths = route => {
+    if (!route) {
+      return [];
+    }
+
+    if (route.views) {
+      return _.castArray(route.views).filter(Boolean).map(absoluteViewPath);
```
gregwolversonHO: If we wanted to remove the reliance on lodash as we go, this could change to:

```
if (route.views) {
  return (Array.isArray(route.views) ? route.views : [route.views])
    .filter(Boolean)
    .map(absoluteViewPath);
}
```

lib/settings.js
Rhodine-orleans-lindsay: I've differentiated them due to the 'watch` and `noCache`properties in the nunjucks configuration. 
`watch: true` automatically reloads templates when files change.
`noCache: true` prevents template caching so edits appear immediately. This is good for local development and debugging.

However, file watching adds unnecessary overhead and caching improves performance and reduces filesystem reads so we would want those properties set to `false` in the testing environments

Branch environments for hof services have the NODE_ENV set to `development` which would set the properties to `true` unnecessarily if I used `development` instead of `local`,  as well as giving a less accurate picture of performance that we should be expecting in uat/stg/prod.

By using `local` instead of `development` you can enable the properties for local development without adversely affecting the branch environment.

gregwolversonHO (approved):

lib/settings.js
Rhodine-orleans-lindsay: Thanks @gregwolversonHO , I have made that change now. I've also made the same change for line 39 and removed lodash from this file

PR #685, opened 2026-06-24
[MINOR] - Vite build task gets output location from config

gregaustinHO:
## What? 
Add the ability to configure the output directory for JavaScript bundle.

## Why?
SAI and Further Subs projects are currently configuring the output directory for JS bundle and this functionality has been removed from newer versions of HOF, so we can not upgrade to new HOF versions.

## How? 
Add config option to builder defaults and pass configuration to Vite build task.

## Check list

### DIFF
--- a/config/builder-defaults.js
+++ b/config/builder-defaults.js
@@ -15,7 +15,8 @@ module.exports = {
     sourceMaps: false
   },
   js: {
-    sourceMaps: false
+    sourceMaps: false,
+    outDir: 'public'
   },
   translate: {
     src: 'apps/**/translations/src',
--- a/README.md
+++ b/README.md
@@ -139,6 +139,16 @@ Each task (except Vite) has a common configuration format with the following opt
 - `match` - defines the pattern for files to watch to trigger a rebuild of this task
 - `restart` - defines if this task should result in a server restart
 
+For Vite, the following configuration can be specified:
+- `outDir` - Specifies the output directory (relative to project root). Only changes only the build output location, does not change runtime static mounting.
+Example from hof settings:
+```js
+  "js": {
+    "outDir": './dist/public'
+  }
+```
+
+
 Additionally the server instance created by `watch` can be configured by setting `server` config. Available options are:
 
 - `cmd` - defines the command used to start the server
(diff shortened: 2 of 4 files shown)


config/builder-defaults.js
PaolaDMadd-Pro, line 1: the initial Browserify settings was"

```
browserify: {
    src: 'assets/js/index.js',
    out: 'public/js/bundle.js',
    match: 'assets/js/**/*.js',
    restart: false,
    compress: false,
    debug: false
  },
```

does any of these properties is valid to keep?

vivekkumar-ho (commented):
Looks good overall. This restores configurability for JS bundle output while keeping default behaviour unchanged, which should be safe for existing services.

A few points to consider:

- We can update readme that js.outDir changes only the build output location, and does not change runtime static mounting.
-  unit tests/coverage for vite config forwarding in both development and production, with default and custom outDir values.

config/builder-defaults.js
gregaustinHO: Hi @PaolaDMadd-Pro . Some of those browserify config options will map more or less onto similar options for Vite, but won't necessarily map exactly. Most of the other options seem concerned with how or when the bundler rebuilds during development. So I don't consider them as important as being able to specify the output directory for the bundle. They could be 'nice to haves' if the team want to add support for them at some point (or any other of the Vite configuration options for that matter). But for now I _think_ the only essential one is the output directory location.

gregaustinHO:
Added update to Readme to cover this option

vivekkumar-ho (approved):

config/builder-defaults.js
PaolaDMadd-Pro: thanks Greg. 
I'll raise the question with the others too. 
@meganjohnHO @dk4g @vivekkumar-ho @anjurajanHO @gregwolversonHO

vivekkumar-ho: We don’t need the old browserify settings because Vite is now responsible for building the JavaScript bundle.

because vite - src: 'assets/js/index.js'

and still outputs:

`public/js/bundle.js`

The only setting we still need from that old config is the output directory:

```
js: {
  outDir: 'public'
}
```
debug - ```
js: {
  sourceMaps: false
}
```

PaolaDMadd-Pro (approved):
lgtm

PR #690, opened 2026-07-01
[MINOR] HOFF-2139 Add always-on prototype pollution protection

vivekkumar-ho:
## What? 

> HOFF-2139

Added always-on prototype pollution protection to the base controller request flow.

The change checks req.body, req.query, and req.params for dangerous prototype pollution keys and values such as:

```
__proto__
prototype
constructor
```
When a suspicious request is detected, processing stops with:

```
400
PROTOTYPE_POLLUTION_DETECTED
```
A security log event is also emitted when request logging is available:

`security.prototype_pollution.blocked`

## Why? 

An incident showed that a crafted URL route param such as:

`/tenant-details/name/__proto__/change`

could flow into form/template handling and trigger server errors, including missing partial template failures.

This change prevents those payloads from reaching form configuration or template rendering, reducing the risk of prototype pollution probes causing application errors or unsafe object behaviour.

## How? 

The protection is implemented centrally in BaseController._configure, which runs before controller-specific form configuration.

- defines a fixed internal blocklist for prototype pollution primitives
- checks request body, query, and route params
- scans nested objects safely
- handles circular objects without crashing
- detects inherited enumerable keys that suggest prototype pollution
- returns a controlled 400 error instead of allowing the request to continue
- cannot be disabled by service-level configuration

## Testing?
- unit tests
- tested using yarn link

## Screenshots (optional)
Screenshot 2026-07-01 at 14 51 24

## Check list

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,10 @@
-## 2026-07-01 Verseion 24.4.0 (Stable), @meganjohnHO
+## 2026-07-01, Version 24.4.0 (Stable), @vivekkumar-ho
+
+### Security
+- Added always-on prototype pollution protection to block dangerous keys and values in request body, query, and route params.
+- Prototype pollution attempts now return `400` with `PROTOTYPE_POLLUTION_DETECTED`.
+
+## 2026-07-01 Version 24.3.2 (Stable), @meganjohnHO
 
 ### Changed
 - Google Analytics domains now added into CSP on presence of `ga4TagId` or `gaTagId` (previously only on `gaTagId`).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### Security
 - Added always-on prototype pollution protection to block dangerous keys and values in request body, query, and route params.
-- Prototype pollution attempts now return `400` with `PROTOTYPE_POLLUTION_DETECTED`.
+- Prototype pollution attempts now return `403` with `PROTOTYPE_POLLUTION_DETECTED`.
 
 ## 2026-07-01 Version 24.3.2 (Stable), @meganjohnHO
 
--- a/controller/base-controller.js
+++ b/controller/base-controller.js
@@ -281,8 +281,18 @@ module.exports = class BaseController extends EventEmitter {
       return;
     }
 
-    if (_.get(this, 'options.logger.warn') && typeof this.options.logger.warn === 'function') {
-      this.options.logger.warn('security.prototype_pollution.blocked', details);
+    const logger = _.get(this, 'options.logger');
+    if (!logger) {
+      return;
+    }
+
+    if (typeof logger.warn === 'function') {
+      logger.warn('security.prototype_pollution.blocked', details);
+      return;
+    }
+
+    if (typeof logger.log === 'function') {
+      logger.log('warn', 'security.prototype_pollution.blocked', details);
     }
   }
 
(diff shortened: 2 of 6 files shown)


controller/base-controller.js
```diff
@@ -206,10 +217,123 @@ module.exports = class BaseController extends EventEmitter {
...
   _configure(req, res, callback) {
     req.form = req.form || {};
     req.form.options = _.cloneDeep(this.options);
-    this.configure(req, res, callback);
+
+    const pollutionDetection = this.checkPrototypePollution(req);
+    if (pollutionDetection) {
+      this.logPrototypePollutionBlocked(req, pollutionDetection);
+      const err = new Error('Prototype pollution detected');
+      err.statusCode = 400;
```
dk4g: I would suggest 403 instead

controller/base-controller.js
```diff
@@ -206,10 +217,123 @@ module.exports = class BaseController extends EventEmitter {
...
+  logPrototypePollutionBlocked(req, details) {
+    if (req && typeof req.log === 'function') {
+      req.log('warn', 'security.prototype_pollution.blocked', details);
+      return;
+    }
+
+    if (_.get(this, 'options.logger.warn') && typeof this.options.logger.warn === 'function') {
+      this.options.logger.warn('security.prototype_pollution.blocked', details);
+    }
+  }
```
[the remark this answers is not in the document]
vivekkumar-ho: valid comment. applied logger fallback

controller/base-controller.js
```diff
@@ -206,10 +217,123 @@ module.exports = class BaseController extends EventEmitter {
...
+        };
+      }
+
+      const blockedValues = new Set(_.get(protection, `blockedValues.${sourceName}`, []));
+      if (this.hasBlockedKeyOrValue(source, blockedKeys, blockedValues)) {
+        return {
+          reason: 'blocked_key_or_value',
+          source: sourceName
+        };
+      }
```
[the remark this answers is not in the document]
vivekkumar-ho: Good catch. added fail-closed handling

controller/base-controller.js
vivekkumar-ho: Updated to 403

vivekkumar-ho:
@meganjohnHO @dk4g @gregwolversonHO I've also included the changes suggested [here](https://github.com/UKHomeOfficeForms/hof/security/code-scanning/14)

CHANGELOG.md
```diff
@@ -1,4 +1,10 @@
-## 2026-07-01 Verseion 24.4.0 (Stable), @meganjohnHO
+## 2026-07-01, Version 24.4.0 (Stable), @vivekkumar-ho
+
+### Security
+- Added always-on prototype pollution protection to block dangerous keys and values in request body, query, and route params.
+- Prototype pollution attempts now return `400` with `PROTOTYPE_POLLUTION_DETECTED`.
```
dk4g:
```suggestion
- Prototype pollution attempts now return `403` with `PROTOTYPE_POLLUTION_DETECTED`.
```

controller/base-controller.js
```diff
@@ -206,10 +217,140 @@ module.exports = class BaseController extends EventEmitter {
...
+
+      try {
+        if (this.hasInheritedEnumerableKey(source)) {
+          return {
+            reason: 'inherited_key',
+            source: sourceName
+          };
+        }
+
+        const blockedValues = new Set(_.get(protection, `blockedValues.${sourceName}`, []));
```
dk4g: just as a simpler alternative without lodash use

```suggestion
        const blockedValues = new Set(protection.blockedValues[sourceName] || []);
```

controller/base-controller.js
```diff
@@ -11,6 +11,17 @@ const ErrorClass = require('./validation-error');
 const Helpers = require('../utilities').helpers;
 const sanitisationBlacklistArray = require('../config/sanitisation-rules');
 
+// Always-on protection against prototype pollution payloads before request data is used by form configuration.
+const defaultPrototypePollutionProtection = {
+  enabled: true,
```
dk4g: I would suggest to remove `enabled: true` it's never used, and misleading

dk4g (changes requested):

controller/base-controller.js
vivekkumar-ho: I think I was planning to add an environment variable initially so we could disable it if needed. However, I realised this is more of a required protection, so having an enable/disable flag does not really make sense. I’ve now removed the enabled key.

controller/base-controller.js
vivekkumar-ho: yeah, make sense. I've removed the usage of lodash from here

CHANGELOG.md
vivekkumar-ho: Updated.

dk4g (approved):

PR #699, opened 2026-08-13
HOFF-1987 remove funkie-phantom and webdriverio

vivekkumar-ho:
## What? 
Removed the vulnerable PhantomJS-based functional test driver stack and migrated the functional test browser runner to Playwright.

This removes:

- `funkie`
- `funkie-phantom`
- `webdriverio`

The functional tests now run directly with Mocha and use headless Chromium through Playwright.

- Also updated dependencies 

## Why? 
Dependabot raised an SSRF alert for `request`. The vulnerable dependency path came through the legacy functional-test stack rather than the application runtime code.

`funkie-phantom` has no safe upgrade path. It is a small wrapper around PhantomJS that starts a WebDriver server on port `4444`, which WebdriverIO then connects to during functional tests. Since the vulnerable dependency could not be resolved by upgrading the existing package, the clean fix was to remove the outdated driver stack.

This also removes reliance on PhantomJS and reduces future maintenance risk around old browser automation dependencies.

## How? 
Replaced the old functional-test flow:

> funkie -> funkie-phantom -> PhantomJS WebDriver server -> WebdriverIO

with:

> Mocha -> Playwright Chromium headless

The existing functional tests were kept largely unchanged. The browser helper now uses Playwright underneath while preserving the small WebdriverIO-style API the tests already use, such as:

- `browser.url()`
- `browser.getUrl()`
- `browser.$()`
- `element.setValue()`
- `element.selectByIndex()`

CI was also updated to install the Playwright Chromium browser binary before running the test suite.

## Testing?
Ran the migrated functional test suite locally:

- yarn run test:functional
- Ran ESLint against the updated functional browser helper:
- GitHub Actions is now green after adding the Playwright browser install step.

## Check list


.github/workflows/automate-tag.yml
```diff
@@ -24,6 +24,7 @@ jobs:
       - uses: borales/actions-yarn@v3.0.0
         with:
           cmd: install --frozen-lockfile
+      - run: yarn playwright install --with-deps chromium
```
dk4g: I would suggest adding a step name, just makes CI logs easier to scan
```suggestion
      - name: Install Playwright Chromium
         run: yarn playwright install --with-deps chromium
```

.github/workflows/automate-tag.yml
```diff
@@ -24,6 +24,7 @@ jobs:
...
       - run: npm test
```
dk4g: I've noticed the line `npm test` which invokes a script that mostly uses `yarn run ...`
it still works fine, it is just a bit inconsistent stylistically

package.json
dk4g: an opportunity to clean this up, `vite` is definitely redundant here

test/functional-tests/lib/browser.js
dk4g, line 1: probably worth adding some comment to explain the purpose of why this file wraps Playwrights's browser and page, and how elements are handled in Browser and Element classes

NOTE: this file acts as a migration compatibility adapter for the subset of the webdriverio API currently used by the project's functional tests. It allows the tests to run on Playwright without immediately rewriting every test

OUT OF SCOPE: Longer term, the tests should probably be refactored toward playwright's native API, especially `page.locator()`, to improve maintainability and make better use of Playwright’s auto-waiting and assertions

.github/workflows/automate-tag.yml
vivekkumar-ho: Yes I agree. Its a good low-risk improvement. code updated

.github/workflows/automate-tag.yml
vivekkumar-ho: make sense. shall we update it to `- run: yarn test` or, for clearer logs we can even add the step name ?

```
- name: Run test suite
  run: yarn test
```

dk4g: agree, that would be helpful!

test/functional-tests/lib/browser.js
vivekkumar-ho: I would agree that this file is doing something intentionally a bit unusual :) I've added comments now and hopefully that would help.

package.json
vivekkumar-ho: Technically, right now, all of these root resolutions appear redundant in the current dependency graph. They are not changing the selected versions anymore. However, I removed vite because it is already a direct dependency at the same range. Perhaps we can review these separately as part of a broader dependency cleanup rather than part of this test-stack migration. The other reason I can think of we can bump some of the major versions but that would need some testing etc instead of just tests specific changes

.github/workflows/automate-tag.yml
vivekkumar-ho: Updated

dk4g (approved):

PR #700, opened 2026-08-13
HOFF-2145 update node and redis versions

dalglishdhandaHO:
## What?
Updated node and Redis versions in testing matrix
## How?
Updated node and redis versions used from the testing matrix.
## Testing?
Tested sandbox
## Screenshots (optional)
image

## Check list

### DIFF
--- a/.github/workflows/automate-publish.yml
+++ b/.github/workflows/automate-publish.yml
@@ -17,7 +17,7 @@ jobs:
       id-token: write
     if: startsWith(github.ref, 'refs/heads/master')
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v7
         with:
           fetch-depth: 0
       - uses: actions/setup-node@v4
--- a/.github/workflows/automate-publish.yml
+++ b/.github/workflows/automate-publish.yml
@@ -17,7 +17,7 @@ jobs:
       id-token: write
     if: startsWith(github.ref, 'refs/heads/master')
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
         with:
           fetch-depth: 0
       - uses: actions/setup-node@v4
--- a/.github/workflows/automate-tag.yml
+++ b/.github/workflows/automate-tag.yml
@@ -10,9 +10,9 @@ jobs:
     strategy:
       matrix:
         node-version: [22.x, 24.x, 26.x]
-        redis-version: [7, 8, 8.4.2]
+        redis-version: [7, 8]
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
       - name: Use Node.js ${{ matrix.node-version }}
         uses: actions/setup-node@v4
         with:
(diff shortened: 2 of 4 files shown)


.github/workflows/automate-tag.yml
```diff
@@ -9,8 +9,8 @@ jobs:
     runs-on: ubuntu-22.04
     strategy:
       matrix:
-        node-version: [20.x, 22.x, 24.x]
-        redis-version: [7, 8]
+        node-version: [24.x, 26.x]
```
vivekkumar-ho: Just to check. Are we removing v22 as well?

dalglishdhandaHO: I believe it has to be removed if package json uses node version higher than what is used in the test matrix's node versions, otherwise there will be failures.

.github/workflows/automate-tag.yml
```diff
@@ -9,8 +9,8 @@ jobs:
...
+        redis-version: [8, 8.4.2]
```
vivekkumar-ho: The requirement read as “add the Redis version currently being used” to the existing 7 and 8 matrix. So I expected Redis 7 to remain unless we are intentionally ending Redis 7 test coverage/support. Is dropping 7 a deliberate support decision ?

dalglishdhandaHO: The ticket isn't very clear as to what it is being asked. I pushed these changes up temporarily with the PR still in draft and the ticket  is not currently in review status.

I see what you mean though, if you think the expectation is to just add 8.4.2 with 7 and 8 for Redis versions

dk4g: I don't think it worth testing with exact `8.4.2` version, testing the latest `8` would be sufficient enough

I agree with @vivekkumar-ho it's not clear why we would want to drop version `7` it's EOL Oct 30, 2027

.github/workflows/automate-tag.yml
dk4g: I would suggest to keep 22 (EOL mid April 2027), unless we are deliberately dropping support for it

.github/workflows/automate-tag.yml
```diff
@@ -9,8 +9,8 @@ jobs:
     runs-on: ubuntu-22.04
     strategy:
       matrix:
-        node-version: [20.x, 22.x, 24.x]
-        redis-version: [7, 8]
+        node-version: [22.x, 24.x, 26.x]
+        redis-version: [7, 8, 8.4.2]
     steps:
       - uses: actions/checkout@v4
```
dk4g: it makes sense to upgrade the checkout action at least to v5 as a part of this PR

> v5.0.0
Update actions checkout to use node 24

https://github.com/actions/checkout/blob/main/CHANGELOG.md#v500

.github/workflows/automate-tag.yml
dalglishdhandaHO: I agree. The ticket does say to check the Redis version being used and add this into the testing matrix too, but I agree it probably wouldn't be needed since we already use 8

.github/workflows/automate-publish.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       id-token: write
     if: startsWith(github.ref, 'refs/heads/master')
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
```
vivekkumar-ho: how about v7 which is the latest?

dk4g (approved):

PR #709, opened 2026-09-09
[PATCH] Removed jQuery from session timeout dialog, sandbox and tests

nzorba:
## What? 
HOFF-1259: Removal of jQuery

## Why? 
Version of jQuery used was outdated

## How? 
- Removed jQuery from session timeout dialog and replaced with native DOM API
- Removed jQuery from frontend tests
- Removed jQuery from sandbox as well as replacing `typeahaed-aria` with `accessible-autocomplete` as there was a conflict between typeahead-aria and its transitive jQuery dependancy causing the country typeahead to not load correctly, and making this switch to access-autocomplete across other services already so the sandbox can be aligned with most services

## Testing?
- Frontend tests updated according to the removal of jQuery 

## Screenshots (optional)
Screenshot 2026-09-09 at 15 54 33
Screenshot 2026-09-09 at 15 55 08

## Check list


frontend/themes/gov-uk/client-js/session-timeout-dialog.js
```diff
@@ -1,40 +1,38 @@
...
-  $lastFocusedEl: null,
-  $closeButton: $('.modal-dialog .js-dialog-close'),
-  $fallBackElement: $('.govuk-timeout-warning-fallback'),
+  lastFocusedEl: null,
+  closeButton: document.querySelector('.modal-dialog .js-dialog-close'),
+  fallBackElement: document.querySelector('.govuk-timeout-warning-fallback'),
   dialogIsOpenClass: 'dialog-is-open',
   timers: [],
-  warningTextPrefix: $('.dialog-text-prefix').text(),
+  warningTextPrefix: document.querySelector('.dialog-text-prefix').textContent,
```
vivekkumar-ho: I’d add a small helper for reading optional DOM text so this module stays defensive when the timeout markup is not present.

At the moment these selectors are dereferenced during module load, so pages/custom layouts without `.dialog-text-prefix` or `.dialog-text` will throw before `init()` can return `false`.

For example:


```
function textContent(selector) {
  const element = document.querySelector(selector);
  return element ? element.textContent : '';
}
```

Then use it here:

```
warningTextPrefix: textContent('.dialog-text-prefix'),
warningText: textContent('.dialog-text'),
```

frontend/themes/gov-uk/client-js/session-timeout-dialog.js
```diff
@@ -1,40 +1,38 @@
...
-  $timer: $('#js-modal-dialog .timer'),
-  $accessibleTimer: $('#js-modal-dialog .at-timer'),
+  timer: document.querySelector('#js-modal-dialog .timer'),
+  accessibleTimer: document.querySelector('#js-modal-dialog .at-timer'),
 
-  secondsSessionTimeout: parseInt($('#js-modal-dialog').data('session-timeout'), 10 || 1800),
-  secondsTimeoutWarning: parseInt($('#js-modal-dialog').data('session-timeout-warning'), 10 || 300),
-  timeoutRedirectUrl: $('#js-modal-dialog').data('url-redirect'),
+  secondsSessionTimeout: parseInt(document.querySelector('#js-modal-dialog').dataset.sessionTimeout, 10 || 1800),
+  secondsTimeoutWarning: parseInt(document.querySelector('#js-modal-dialog').dataset.sessionTimeoutWarning, 10 || 300),
```
vivekkumar-ho: The `dialogDatasetValue` (reference below) helper makes the dataset lookup safe, but the fallback here is not applied as intended. `10 || 1800` and `10 || 300` both evaluate to `10`, so missing values will still parse as `NaN`.

Could we apply the default after parsing instead?

frontend/themes/gov-uk/client-js/session-timeout-dialog.js
```diff
@@ -1,40 +1,38 @@
...
+  timeoutRedirectUrl: document.querySelector('#js-modal-dialog').dataset.urlRedirect,
```
vivekkumar-ho: Could we add a small parsing helper to handle the null undefined values?

```
function dialogDatasetValue(key) {
  const element = document.getElementById('js-modal-dialog');
  return element ? element.dataset[key] : undefined;
}
```

frontend/themes/gov-uk/client-js/session-timeout-dialog.js
```diff
@@ -277,8 +277,8 @@ window.GOVUK.sessionDialog = {
   },
 
   refreshSession: function () {
-    $.get('')
-      .done(function () {
+    fetch('')
+      .then(function () {
```
vivekkumar-ho: Since this is replacing `$.get().done(...)` with `fetch(...)`, we need the `response.ok` check here. `fetch` still resolves `.then(...)` for HTTP error responses, while the old `.done(...)` callback only ran for successful responses.

Without this check, a failed keep-alive response could still update `timeSessionRefreshed` and restart the timeout controller.

test/frontend/jest/sessionDialog.test.js
```diff
@@ -7,10 +7,7 @@ const sessionTimeoutWarningHtml = fs.readFileSync(path.resolve(__dirname, '../..
 jest.dontMock('fs');
```
vivekkumar-ho, line 7: should update tests according to above review comments

vivekkumar-ho (changes requested):
I think we should take this opportunity to upgrade packages to the latest compatible versions in both HOF and the sandbox. That would help us address a number of the existing vulnerabilities and dependabot alerts at the same time, rather than having to revisit them separately later.

dalglishdhandaHO (approved):

sandbox/package.json
```diff
@@ -15,10 +15,9 @@
   },
   "author": "",
   "dependencies": {
+    "accessible-autocomplete": "^3.0.1",
     "govuk-frontend": "4.10",
-    "jquery": "^3.7.1",
-    "sass": "^1.101.0",
-    "typeahead-aria": "^1.0.4"
+    "sass": "^1.101.0"
```
vivekkumar-ho, line 20:
```suggestion
    "sass": "^1.104.0"
```

This will resolve a few dependabot alerts

sandbox/package.json
vivekkumar-ho, line 25: this can be reviewed. shall we remove it instead ?

frontend/themes/gov-uk/client-js/session-timeout-dialog.js
```diff
@@ -1,40 +1,50 @@
...
   // Timer specific markup. If these are not present, timeout and redirection are disabled
-  $timer: $('#js-modal-dialog .timer'),
-  $accessibleTimer: $('#js-modal-dialog .at-timer'),
+  timer: document.querySelector('#js-modal-dialog') ? document.querySelector('#js-modal-dialog .timer') : null,
+  accessibleTimer: document.querySelector('#js-modal-dialog') ? document.querySelector('#js-modal-dialog .at-timer') : null,
 
-  secondsSessionTimeout: parseInt($('#js-modal-dialog').data('session-timeout'), 10 || 1800),
-  secondsTimeoutWarning: parseInt($('#js-modal-dialog').data('session-timeout-warning'), 10 || 300),
-  timeoutRedirectUrl: $('#js-modal-dialog').data('url-redirect'),
+  secondsSessionTimeout: parseInt(dialogDatasetValue('sessionTimeout'), 10 || 1800),
```
vivekkumar-ho, line 33: The defensive dataset lookup is good, but the fallback here still technically is not applied. 10 || 1800 always evaluates to 10, so missing/invalid values still become NaN. Could we apply the default after parsing instead?

frontend/themes/gov-uk/client-js/session-timeout-dialog.js
```diff
@@ -1,40 +1,50 @@
...
+  secondsTimeoutWarning: parseInt(dialogDatasetValue('sessionTimeoutWarning'), 10 || 300),
```
vivekkumar-ho, line 34: The defensive dataset lookup is good, but the fallback here still technically is not applied. 10 || 300 always evaluates to 10, so missing/invalid values still become NaN. Could we apply the default after parsing instead?

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 2026-09-09 Version 24.5.1 (Stable), @nzorba
+
+### Changed
+- Removed jQuery from the session timeout dialog and replaced its DOM operations with native browser APIs.
+- Removed the jQuery dependency and updated frontend tests to use native DOM APIs and `fetch`.
+
```
vivekkumar-ho, line 6: Based on which branch gets merged first, please update the changelog accordingly. 

https://github.com/UKHomeOfficeForms/hof/pull/710/changes#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR1-R9

sandbox/package.json
nzorba: @vivekkumar-ho are you referring to the nodemon?

vivekkumar-ho: @nzorba selective version resolutions. 

```
  "resolutions": {
    "picomatch": "^2.3.2"
  }
```

dalglishdhandaHO (approved):
LGTM. Should discuss if/how to include days.js migration work in the same patch as this and how to proceed before merging in

vivekkumar-ho (approved):
Looks fine to me. Whether we use the beta version or the updated patch version, I’d still like to test the following scenarios to ensure that the existing functionality hasn’t been affected.

- Session timeout happy path: confirm the warning dialog appears, countdown updates, “Stay signed in / Stay on this page” closes the dialog, refreshes the session, and leaves the user on the same page.

- Session timeout expiry path: allow the timer to expire and confirm the user is redirected to /session-timeout with the expected service behavior.

- Custom session timeout content: if the service customises timeout text, translations, templates, save-and-exit, or exit links, confirm those still render and route correctly.

- Accessibility and keyboard checks: confirm focus returns after closing the timeout dialog, Esc closes it, background content is not reachable while open, and live-region/countdown text behaves as expected.

- No-JS fallback: confirm the fallback timeout warning and eventual timeout/redirect still work when JavaScript is disabled.

dalglishdhandaHO (approved):

PR #711, opened 2026-09-11
HOFF-1999 migrate from momentjs to dayjs

dalglishdhandaHO:
## What? 
Migrate from momentjs to dayjs and updated tests to match
## Testing?
Made a submission in sandbox
## Screenshots (optional)
image

## Check list


dalglishdhandaHO:
TO DO: Add to changelog after other PR for removing jQuery has gone through

middleware/rate-limiter.js
```diff
@@ -74,9 +74,9 @@ module.exports = (options, rateLimitType) => {
         }
         // if number of requests made is less than allowed maximum, log new entry
         const lastRequestLog = data[data.length - 1];
-        const potentialCurrentWindowIntervalStartTimeStamp = currentRequestTime
-          .subtract(WINDOW_LOG_INTERVAL_IN_MINUTES, 'minutes')
-          .unix();
+        currentRequestTime = currentRequestTime
```
MikeMaratheftis: Do we need to update `currentRequestTime` here? Could we keep this as a single statement like the original implementation, just using the Day.js syntax?

dalglishdhandaHO: Good spot. Even for a readability standpoint, it makes more sense to keep it all together.

dalglishdhandaHO: Apologies, I neglected to see another change I had earlier in the file which would show why this would not work.
With currentRequestTime no longer being a const, it can undergo reassignment, which is why the way this was done is very slightly different.
As for why we are using a let instead of a const, this is to adjust with how a mutation was happening with moment.js before.
Therefore, there is a very slight difference between what is being used now compared to before, and that is enough to cause a failure with rate-limiter

nzorba (approved):
LGTM!

